PR #22, opened 2015-04-22
Add IDockerClient interface

ryanwentzel:
Add `IDockerClient` interface as discussed in https://github.com/ahmetalpbalkan/Docker.DotNet/issues/21.


ahmetalpbalkan:
LGTM. @jgarverick what do you think?

jgarverick:
LGTM as well. You may fire when ready. ;)

Issue #34, opened 2015-08-31
Docker Exec


rjperes:
Hi,
Am I right to assume that right now it isn't possible to use the Docker EXEC command with Docker.DotNet?
Thanks!
RP

ahmetalpbalkan:
Sounds about right. Implementation should be rather straightforward. Please consider contributing.

rjperes:
I think I will!

rjperes:
I just submitted a pull request: https://github.com/ahmetalpbalkan/Docker.DotNet/pull/35.
Waiting for your feedback!

skarred14:
Hi, DockerDotNet API has really been useful to me for creating applications based on infrastructural changes and processes - thank you for such a string utility kit. Particularly, in reference to "ExecCreateContainerAsync", the method returns the ExecCreateContainerResponse object. However, I do not see the responsive action being executed. Here are the Config parameters that I entered: Config = new ExecCreateContainerConfig() {
                                                                                                    AttachStdin = true,
                                                                                                    AttachStdout = true,
                                                                                                    AttachStderr = true,
                                                                                                    Tty = true,
                                                                                                    Cmd = new List<string>() {
                                                                                                        "mkdir" , "abc"  }
                                                                                                }.
And here is the function call,
ExecCreateContainerResponse ex = await client.Containers.ExecCreateContainerAsync(cr.Id, new ExecCreateContainerParameters() { Config = ...<as given above>}
    );
where, cr.Id is the ID of the container that I have been able to create and run using CreateContainerAsync and StartContainerAsync. 
I expected that the outcome of the above instruction would result in the creation of a new directory "abc" within the container instance. Although the program executed successfully providing a valid ExecCreateContainerResponse (Id, Warnings), when I looked up, by attaching the container on console, I couldnt find the directory. Could you please suggest, as to how this can be achieved - I am trying to get to a result similar to "docker exec -d containerName mkdir abc" on console prompt.

ahmetalpbalkan:
@skarred14 please open another separate issue about your problem. This originally reported ossue is apparently resolved with a submitted patch.

PR #35, opened 2015-09-02
Implementation of issue #34: Docker Exec

rjperes:


Docker.DotNet/ContainerOperations.cs
```diff
@@ -134,6 +134,23 @@ public Task<Stream> ExportContainerAsync(string id, CancellationToken cancellati
             return response.StatusCode != HttpStatusCode.NotModified;
         }
 
+        public async Task<ExecContainerResponse> ExecContainerAsync(string id, Config config)
```
ahmetalpbalkan: instead of directly passing in config, please create a `ExecContainerParameters` and have Config as a field in there (see rest of the file for examples)

rjperes:
But there are also examples where the id is passed along with a config parameter.

Docker.DotNet/ContainerOperations.cs
```diff
@@ -134,6 +134,23 @@ public Task<Stream> ExportContainerAsync(string id, CancellationToken cancellati
...
+        {
+            if (string.IsNullOrEmpty(id))
+            {
+                throw new ArgumentNullException("id");
+            }
+
+            string path = string.Format(CultureInfo.InvariantCulture, "containers/{0}/exec", id);
+            JsonRequestContent<Config> data = null;
```
ahmetalpbalkan: also `/exec` does not take a Config object... It accepts only a small number of fields:

http://docs.docker.com/reference/api/docker_remote_api_v1.20/#exec-create

Please create a `ExecContainerParameters` object with these fields.

ahmetalpbalkan:
@rjperes thanks for the pull request. I left some comments, if you could address those the rest is LGTM.

rjperes:
?
The fields accepted by Exec are exactly the same as the Config class contains!

ahmetalpbalkan:
@rjperes please leave the comment below the related comment :)

ahmetalpbalkan:
@rjperes I didn't say you should move `id` to ExecParameters. 

    public async Task<ContainerProcessesResponse> ListProcessesAsync(string id, ListProcessesParameters parameters)
    public Task KillContainerAsync(string id, KillContainerParameters parameters)

do you see a direct input object like Config here? You need to create ExecContainerParameters with a single field: Config.

ahmetalpbalkan:
> The fields accepted by Exec are exactly the same as the Config class contains!

No they are not. 
* Config class has 20+ memebers.  https://github.com/ahmetalpbalkan/Docker.DotNet/blob/master/Docker.DotNet/Models/Config.cs 
* `/exec` has 5. http://docs.docker.com/reference/api/docker_remote_api_v1.20/#exec-create

Docker.DotNet/IContainerOperations.cs
```diff
@@ -20,6 +20,8 @@ public interface IContainerOperations
 
         Task<Stream> ExportContainerAsync(string id, CancellationToken cancellationToken);
 
+        Task<ExecContainerResponse> ExecContainerAsync(string id, ExecContainerParameters parameters);
```
ahmetalpbalkan: thanks for updating the PR. I realized there are a lot of Exec operations (ExecCreate, ExecStart, ExecResize etc...) also ExecContainer is not the right terminology. Can we rename this param (and ...Parameters type) to `ContainerExecCreate` please?

Docker.DotNet/Models/ExecConfig.cs
```diff
@@ -0,0 +1,23 @@
...
+        public bool AttachStdin { get; set; }
+
+        [DataMember(Name = "AttachStdout")]
+        public bool AttachStdout { get; set; }
+
+        [DataMember(Name = "AttachStderr")]
+        public bool AttachStderr { get; set; }
+
+        [DataMember(Name = "Tty")]
+        public bool Tty { get; set; }
```
ahmetalpbalkan: for instance here if I don't send any value to this constructor, the value will be false.... However maybe in the actual REST call if I omit this field, it could be defaulting the `true` but here you're making the default `false`... :) In order to avoid that, please use `bool?` instead of `bool` and let users transmit this request without specifying any value for these bool fields.

Models directory is full of examples of this: https://github.com/ahmetalpbalkan/Docker.DotNet/blob/c2ca24da15ba1c12be8552ef52a58f07a111351f/Docker.DotNet/Models/RemoveContainerParameters.cs

rjperes: Right.

rjperes:
BTW, why not make Docker.DotNet.BasicAuth and Docker.DotNet.X509 also portable libraries? I'm not sure X509 could be made portable, but BasicAuth certainly could. Not that important, though.

ahmetalpbalkan:
> BTW, why not make Docker.DotNet.BasicAuth and Docker.DotNet.X509 also portable libraries? I'm not sure X509 could be made portable, but BasicAuth certainly could. Not that important, though.

please open a separate issue for discussion... Also please don't @[mention] people in the commit messages, I'm getting extra emails because of that.

rjperes:
Of course it compiles... spaces are optional.

ahmetalpbalkan:
> Of course it compiles... spaces are optional.

but I'm pretty sure that's not the language convention. I've been working at Microsoft for 3 years and that's the first time I've ever seen such a thing (meaning, it has a surprise factor). Does not your IDE fix that when you auto-format the code?

Docker.DotNet/ContainerOperations.cs
```diff
@@ -134,6 +134,23 @@ public Task<Stream> ExportContainerAsync(string id, CancellationToken cancellati
             return response.StatusCode != HttpStatusCode.NotModified;
         }
 
+        public async Task<ContainerExecCreateResponse> ExecContainerAsync(string id, ContainerExecCreateParameters parameters)
```
ahmetalpbalkan: I think you forgot to rename this method.

rjperes: OK.

rjperes:
Are you going to merge this?

ahmetalpbalkan:
Sorry entirely forgot about this. Merging now.

rjperes:
Thanks! And what about the NuGet package?

ahmetalpbalkan:
@rjperes just pushed 1.2.0 to NuGet.

PR #40, opened 2015-11-24
made DockerClient to use only one handler per instance

tugberkugurlu:
The current implementation of `DockerClient` does two things wrong:

 - It creates an `HttpClient` instance each time you make a request. This will result in creating inner handlers in each request which does the connection pooling. 
 - It leaves `HttpClient` instances undisposed which will result in open TCP connections as inner handlers don't close them for certain period of time by default.

The best usage of `HttpClient` is to create one instance of this and use it for the entire app lifetime as its members are thread safe + its inner handler does connection pooling. See [here](http://stackoverflow.com/questions/15705092/do-httpclient-and-httpclienthandler-have-to-be-disposed) for more info.

However, as you have special cases for credentials, I started disposing `HttpClient` but left the inner handler undisposed for a `DockerClient` instance (this might be handled better but I thought this would be enough as the first pass on this). Disposing `DockerClient` instance will dispose everything. Having one instance of `DockerClient` per application lifecycle would be the recommended usage for `DockerClient` now.

Didn't have a chance to do a manual run to see whether I screwed up anything. So, it would be good to go through that first. I can probably do that later sometimes if you are OK to wait a bit more on this.

### DIFF
--- a/Docker.DotNet.BasicAuth/BasicAuthCredentials.cs
+++ b/Docker.DotNet.BasicAuth/BasicAuthCredentials.cs
@@ -8,6 +8,7 @@ namespace Docker.DotNet.BasicAuth
 {
     public class BasicAuthCredentials : Credentials
     {
+        private readonly HttpMessageHandler _handler;
         private readonly SecureString _username;
         private readonly SecureString _password;
         private readonly bool _isTls;
@@ -24,6 +25,7 @@ namespace Docker.DotNet.BasicAuth
                 throw new ArgumentException("password");
             }
 
+            _handler = new HttpClientHandler();
             _username = username;
             _password = password;
             _isTls = isTls;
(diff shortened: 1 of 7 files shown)


ahmetalpbalkan:
@tugberkugurlu thanks a bunch. Generally looks good to me. I was not aware of the behavior at all. As you said we probably should test this out manually. Would you like to sign up for that or prefer me to do?

tugberkugurlu:
@ahmetalpbalkan would be good to have someone else's eyes on this rather than mine. I would be free to do that this weekend though if you don't have time.

ahmetalpbalkan:
I'll try.

ahmetalpbalkan:
I tried this with unauthenticated client mode. Seems like it's working fine. LGTM.

ahmetalpbalkan:
This went out in release `v1.2.1`, thanks.

tugberkugurlu:
:sparkles:

PR #42, opened 2015-12-23
Fix for defect where streams are being closed prematurely

ItsVeryWindy:
When calling methods that return a stream, the stream cannot be read from due to HttpClient being disposed, which results in any requests still in flight being disposed of.

I've done some refactoring so that HttpClient has only one instance and then the credential instances inject a httpmessagehandler rather than being responsible for creating the HttpClient.

Because I've had to make it a single instance I've had to make changes to the way timeouts work, so instead of setting the timeout value on HttpClient I configure the cancellation token to timeout which behaviour wise is the same.

I've also added configure await on awaited calls just as it's considered good practice.

### DIFF
--- a/Docker.DotNet.BasicAuth/BasicAuthCredentials.cs
+++ b/Docker.DotNet.BasicAuth/BasicAuthCredentials.cs
@@ -1,11 +1,6 @@
 ﻿using System;
 using System.Net.Http;
-using System.Net.Http.Headers;
-using System.Runtime.InteropServices;
 using System.Security;
-using System.Text;
-using System.Threading;
-using System.Threading.Tasks;
 
 namespace Docker.DotNet.BasicAuth
 {
@@ -13,47 +8,6 @@ namespace Docker.DotNet.BasicAuth
     {
         private readonly BasicAuthHandler _handler;
         private readonly bool _isTls;
-
-        private class BasicAuthHandler : HttpClientHandler
-        {
-            private readonly SecureString _username;
-            private readonly SecureString _password;
-
-            public BasicAuthHandler(SecureString username, SecureString password)
-            {
-                _username = username;
-                _password = password;
-            }
-
-            protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
-            {
-                request.Headers.Authorization = new AuthenticationHeaderValue("Basic", BuildParameters());
-
-                return base.SendAsync(request, cancellationToken);
-            }
-
-            private string BuildParameters()
-            {
-                string authInfo = string.Format("{0}:{1}", ConvertToUnsecureString(_username),
-                    ConvertToUnsecureString(_password));
-
-                return Convert.ToBase64String(Encoding.Default.GetBytes(authInfo));
-            }
-
-            private static string ConvertToUnsecureString(SecureString secureString)
-            {
-                IntPtr unmanagedString = IntPtr.Zero;
-                try
-                {
-                    unmanagedString = Marshal.SecureStringToGlobalAllocUnicode(secureString);
-                    return Marshal.PtrToStringUni(unmanagedString);
-                }
-                finally
-                {
-                    Marshal.ZeroFreeGlobalAllocUnicode(unmanagedString);
-                }
-            }
-        }
 
         public override HttpMessageHandler Handler
         {
@@ -75,7 +29,7 @@ namespace Docker.DotNet.BasicAuth
                 throw new ArgumentException("password");
             }
 
-            _handler = new BasicAuthHandler(username, password);
+            _handler = CreateHandler(username, password);
 
             _isTls = isTls;
         }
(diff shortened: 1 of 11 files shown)


ahmetalpbalkan:
@ItsVeryWindy Hello, thanks a lot for this pull request. It looks good to me overall. I'd love to hear @jgarverick's opinion as well. I don't have any windows build environment nowadays, does your change break the PCL build?

Also, are the changes in `Docker.DotNet.BasicAuth/BasicAuthCredentials.cs` relevant to HttpClient-related changes? If not, let's take it in a separate PR.

ItsVeryWindy:
It builds just fine, I've not worked on PCL libraries so I'm not sure if there are additional steps required to make sure.

The changes to BasicAuthCredentials are relevant, because I've shifted the responsibility of managing the HttpClient to the DockerClient instead of it being managed by the Credentials class. Credentials now returns a HttpMessageHandler instead of creating a httpclient so I've had to change the way the authorization headers were being added.

Docker.DotNet/DockerClient.cs
```diff
@@ -74,31 +80,50 @@ internal Task<Stream> MakeRequestForStreamAsync(IEnumerable<ApiResponseErrorHand
...
+            return await response.Content.ReadAsStreamAsync().ConfigureAwait(false);
         }
 
-        private async Task<HttpResponseMessage> MakeRequestInnerAsync(TimeSpan? requestTimeout, HttpCompletionOption completionOption, HttpMethod method, string path, IQueryString queryString, IDictionary<string, string> headers, IRequestContent data, CancellationToken cancellationToken)
+        private Task<HttpResponseMessage> MakeRequestInnerAsync(TimeSpan? requestTimeout, HttpCompletionOption completionOption, HttpMethod method, string path, IQueryString queryString, IDictionary<string, string> headers, IRequestContent data, CancellationToken cancellationToken)
         {
-            using (HttpClient client = this.GetHttpClient())
+            HttpRequestMessage request = PrepareRequest(method, path, queryString, headers, data);
+
+            if (requestTimeout.HasValue)
```
ahmetalpbalkan: can you please make this `requestTimeout.HasValue && requestTimeout.Value != InfiniteTimeout`

Docker.DotNet.BasicAuth/BasicAuthCredentials.cs
```diff
@@ -1,18 +1,68 @@
...
+                    return Marshal.PtrToStringUni(unmanagedString);
+                }
+                finally
+                {
+                    Marshal.ZeroFreeGlobalAllocUnicode(unmanagedString);
+                }
+            }
+        }
+
+        private readonly BasicAuthHandler _handler;
```
ahmetalpbalkan: do you mind moving these private field declarations all the way up next to _username & _password?

ahmetalpbalkan: do you mind grouping private member declarations all the way up in the type declaration, like it was before the change?

Docker.DotNet.BasicAuth/BasicAuthCredentials.cs
```diff
@@ -1,18 +1,68 @@
...
+using System.Threading.Tasks;
 
 namespace Docker.DotNet.BasicAuth
 {
     public class BasicAuthCredentials : Credentials
     {
-        private readonly HttpMessageHandler _handler;
-        private readonly SecureString _username;
-        private readonly SecureString _password;
+        private class BasicAuthHandler : HttpClientHandler
```
ahmetalpbalkan: any specific reason why we're embedding this `BasicAuthHandler ` class inside `BasicAuthCredentials` class? If not, let's move out.

Docker.DotNet/DockerClient.cs
ItsVeryWindy: If I merged those conditions on to one line it would execute the else statement if the value was infinite timeout which would result in the timeout being set to the default timeout for HttpClient, ie (100 seconds)

ahmetalpbalkan: @ItsVeryWindy ah you're right. and I bet `CreateTimeoutToken(InfiniteTimeout)` is not a good idea either. Looks good then.

Docker.DotNet.BasicAuth/BasicAuthCredentials.cs
ItsVeryWindy: I just wanted to keep the functionality internal to that class, there is no technical reason why that couldn't be moved out if that is desired.

Docker.DotNet.BasicAuth/BasicAuthCredentials.cs
ItsVeryWindy: Done.

ItsVeryWindy:
I've made some additional changes to clean things up and ended up moving BasicAuthHandler out of BasicAuthCredentials which would make it more testable.

jgarverick:
LGTM as well.  Since the basic auth module is outside the scope of the PCL library it shouldn't be an issue.  It does build just fine for me.  I'll take a deeper look later, though, and try to wire it up to a UWP app or something, just to be certain.

ahmetalpbalkan:
Thanks a lot @jgarverick @itsverywindy.

PR #43, opened 2015-12-30
Added features introduced in api version 1.20

ItsVeryWindy:
This is to introduce api changes introduced in 1.20.
* `GET /containers/(id)/archive` get an archive of filesystem content from a container.
* `PUT /containers/(id)/archive` upload an archive of content to be extracted to an existing directory inside a container’s filesystem.
* `POST /containers/(id)/copy` is deprecated in favor of the above `archive` endpoint which can be used to download files and directories from a container.
* The `hostConfig` option now accepts the field `GroupAdd`, which specifies a list of additional groups that the container process will run as.

Most of it is pretty similar to existing methods, I had to add an extra method to dockerclient so I could retrieve the stream along with its headers. One thing I wasn't quote sure about was whether the HEAD request to the archive endpoint should be a separate method, the difference between GET and HEAD is that HEAD won't return a stream.

### DIFF
--- a/Docker.DotNet/IContainerOperations.cs
+++ b/Docker.DotNet/IContainerOperations.cs
@@ -42,7 +42,7 @@ namespace Docker.DotNet
 
         Task<Stream> CopyFromContainerAsync(string id, CopyFromContainerParameters parameters, CancellationToken cancellationToken);
 
-        Task<GetArchiveFromContainerResponse> GetArchiveFromContainerAsync(string id, GetArchiveFromContainerParameters parameters, bool statsOnly, CancellationToken cancellationToken);
+        Task<GetArchiveFromContainerResponse> GetArchiveFromContainerAsync(string id, GetArchiveFromContainerParameters parameters, bool statOnly, CancellationToken cancellationToken);
 
         Task ExtractArchiveToContainerAsync(string id, ExtractArchiveToContainerParameters parameters, Stream stream, CancellationToken cancellationToken);
     }
--- a/Docker.DotNet/IContainerOperations.cs
+++ b/Docker.DotNet/IContainerOperations.cs
@@ -41,5 +41,9 @@ namespace Docker.DotNet
         Task<Stream> GetContainerLogsAsync(string id, GetContainerLogsParameters parameters, CancellationToken cancellationToken);
 
         Task<Stream> CopyFromContainerAsync(string id, CopyFromContainerParameters parameters, CancellationToken cancellationToken);
+
+        Task<GetArchiveFromContainerResponse> GetArchiveFromContainerAsync(string id, GetArchiveFromContainerParameters parameters, bool statsOnly, CancellationToken cancellationToken);
+
+        Task ExtractArchiveToContainerAsync(string id, ExtractArchiveToContainerParameters parameters, Stream stream, CancellationToken cancellationToken);
     }
 }
--- a/Docker.DotNet/Models/ContainerPathStats.cs
+++ b/Docker.DotNet/Models/ContainerPathStats.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.Runtime.Serialization;
+
+namespace Docker.DotNet.Models
+{
+    [DataContract]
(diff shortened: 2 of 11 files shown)


Docker.DotNet/IContainerOperations.cs
```diff
@@ -41,5 +41,9 @@ public interface IContainerOperations
         Task<Stream> GetContainerLogsAsync(string id, GetContainerLogsParameters parameters, CancellationToken cancellationToken);
 
         Task<Stream> CopyFromContainerAsync(string id, CopyFromContainerParameters parameters, CancellationToken cancellationToken);
+
+        Task<GetArchiveFromContainerResponse> GetArchiveFromContainerAsync(string id, GetArchiveFromContainerParameters parameters, bool statsOnly, CancellationToken cancellationToken);
```
ahmetalpbalkan: I think we should call it `stat` here, not `stats`. It refers to the stat system call in posix and not "statistics" which "stats" usually refers to. Also there is a stats `GET /containers/(id)/stats` endpoint, which is a different thing, but if we use "stats" in this method, it'll be confusing.

Docker.DotNet/Models/ContainerPathStats.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.Runtime.Serialization;
+
+namespace Docker.DotNet.Models
+{
+    [DataContract]
+    public class ContainerPathStats
```
ahmetalpbalkan: Follow up to my stats vs stat comment, we can probably call this ContainerPathInfo or ContainerPathStat.

Docker.DotNet/Models/ExtractArchiveToContainerParameters.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Docker.DotNet.Models
+{
+    public class ExtractArchiveToContainerParameters
+    {
+        [QueryStringParameter("path", true)]
+        public string Path { get; set; }
+
+        [QueryStringParameter("noOverwriteDirNonDir", false, typeof(BoolQueryStringConverter))]
+        public bool NoOverwriteDirNonDir { get; set; }
```
ahmetalpbalkan: It appears like this argument is not required. Can we make this field `bool?` type so that it's omitted unless explicitly set by the user?

Docker.DotNet/Models/HostConfig.cs
```diff
@@ -12,6 +12,9 @@ public class HostConfig
         [DataMember(Name = "Links")]
         public IList<string> Links { get; set; }
 
+        [DataMember(Name = "GroupAdd")]
+        public IList<string> GroupAdd { get; set; }
```
ahmetalpbalkan: I can't seem to find GroupAdd in https://docs.docker.com/engine/reference/api/docker_remote_api_v1.20/ ?

ahmetalpbalkan:
@ItsVeryWindy thanks a bunch for this pull request, it's marvelous. Generally it looks good to me. If you could address the issues and ping me again, I'll try to merge it as soon as possible and release a new version. Since we're not fully catching up with API version 1.20, docs will still indicate we support 1.16.

Docker.DotNet/Models/HostConfig.cs
ItsVeryWindy: Yeah, I found it unusual that it wasn't listed in the docs themselves and only in the api changes. I had to check the source code to confirm its existence and to determine what type it was.

ahmetalpbalkan: Hmm alright then, I'm guessing you really need this field. :) Also please consider making a docs PR to Docker.

ItsVeryWindy:
Issues have been addressed.

ahmetalpbalkan:
@ItsVeryWindy thanks a bunch! merging. I'll be releasing a new version now as well.

ahmetalpbalkan:
Release is out https://www.nuget.org/packages/Docker.DotNet/1.2.2 thanks again.

PR #55, opened 2016-04-28
Temporarily removing support for netstandard1.3

swernli:
@jterry75 @ahmetalpbalkan @jstarks 
We've hit an issue compiling for netstandard1.3 using the latest version of the dotnet CLI (reported here: https://github.com/dotnet/cli/issues/2739). The current mechanism used to incorrectly report success during compile, but required runtime references couldn't be found with the latest dotnet CLI.  Until we have a resolution to the issue or better guidance on how to compile for DNXCore, it's safer in the short term to remove that build target.

### DIFF
--- a/Docker.DotNet.BasicAuth/project.json
+++ b/Docker.DotNet.BasicAuth/project.json
@@ -8,22 +8,6 @@
     "win10-x86": {}
   },
   "frameworks": {
-    "netstandard1.3": {
-      "imports": [
-        "dotnet",
-        "netcore50",
-        "uap10"
-      ],
-      "compilationOptions": {
-        "define": [
-          "netstandard"
-        ]
-      },
-      "dependencies": {
-        "System.Runtime": "4.0.20",
-        "System.Text.Encoding": "4.0.10"
-      }
-    },
     "net46": {
       "frameworkAssemblies": {
         "System.Runtime": "",
--- a/Docker.DotNet/project.json
+++ b/Docker.DotNet/project.json
@@ -8,28 +8,6 @@
     "win10-x86": {}
   },
   "frameworks": {
-    "netstandard1.3": {
-      "imports": [
-        "dotnet",
-        "netcore50",
-        "uap10"
-      ],
-      "dependencies": {
-        "System.Collections": "4.0.10",
-        "System.IO": "4.0.10",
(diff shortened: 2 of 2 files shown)


swernli:
I confirmed that this restores, builds and publishes as expected with the latest dotnet CLI.

ericstj:
How is this related to the referenced issue?  You're talking apples and oranges here `nestandard1.3` is different from `netcore50` referenced in https://github.com/dotnet/cli/issues/2739.

swernli:
@ericstj There is a lot of confusion here on our part, but in the short term we just need to make sure this project keeps compiling.  

Without going into too much detail (I'll leave that to @jterry75, who has a better grasp of the problem than I), the project.json as it exists in master today seemed to work fine for dotnet CLI beta (1.0.0-beta-002252), but when we recently updated to dotnet CLI rc2 (1.0.0-rc2-002533) it started failing during restore because of dependencies we haven't been expressing correctly.  We've tried several approaches to explicitly enumerate the separate dependencies for netcore50, but none of them allowed us to successfully publish targeting the runtimes we needed.  Issue https://github.com/dotnet/cli/issues/2739 is us asking for help on finding out the right way to get this project to compile for core, but in the meantime we need to unblock compilation with dotnet rc2, and temporarily removing the netstandard1.3 section in favor of relying on the net45 and net46 sections of the project.json is the simplest way to unblock us.

Once we figure out (hopefully with your help) the right way to compile this, we can get our project into shape!

jstarks:
LGTM, assuming we follow up with the right fix soon

ericstj:
Sure, let me know what type of library you are trying to build.  `netcore50` is for UWP apps and I wouldn't expect this showing up in any projects built with the CLI, with the one exception being a *library* which needs to cross-compile for UWP for some reason (eg: to avoid PInvokes that are available in UWP).

swernli:
@ericstj Thanks Eric!  It looks like @jterry75 may have figured out the trick for getting this to compile the way we want.  He hasn't made a PR yet, but you can see the commit on his branch here: https://github.com/jterry75/Docker.DotNet/commit/cd73be5228998c984bbb1dc4f3074c2f05b8245a.  Would you mind taking a look and providing any feedback?

Issue #56, opened 2016-04-29
Project should compile for netcore50 using latest dotnet CLI


swernli:
The project.json file isn't in the right format to allow publishing for netcore50 on our desired runtimes with the latest dotnet CLI.  https://github.com/ahmetalpbalkan/Docker.DotNet/pull/55 temporarily removes support for netstandard1.3 as a work-around, but long term we'd like to have the correct project.json to support netcore properly.

jterry75:
#57

jterry75:
This wasnt fixed correctly after recent changes.

jterry75:
New PR on its way.

PR #57, opened 2016-04-29
Fixes the build for .NET Core,uap10 in the netstandard moniker.

jterry75:


jstarks:
LGTM, is this ready to go?

jterry75:
We were waiting for some feedback from dotnet CLI #56 but this does work. I think this is the write refactor here.

swernli:
LGTM.

Issue #59, opened 2016-05-03
Some of the model types for int/uint should be long/ulong


jterry75:
In Go int/uint are compiled to the runtime size. Because this is likely 64bit the int/uint generated code is the wrong size and will fail in the C# models. They should be generated as long/ulong instead.

PR #60, opened 2016-05-03
Resolves: #59

jterry75:
1. Fixes a bug where in Go int and uint are converted to the runtime type
size not the standard 32bit like in C#. This updates the generator to
output int and uint as long and ulong instead so we dont get type
mismatches from the service.


jterry75:
@swernli - FYI

Referenced from: https://github.com/Microsoft/Docker-PowerShell/issues/39

jstarks:
LGTM

Issue #47, opened 2016-02-26
X509 not suport DNXcore


mateusdemboski:
I trying install the library Docker.DotNet.X509 on my ASP.NET Core project, but VS it's displaying the mensage 'NU1002 The dependency Docker.DotNet.X509 1.2.2 in project <ProjectName> does not support framework DNXCore,Version=v5.0.'

How i can fix this? or is a bug?

Regards

ahmetalpbalkan:
The package probably doesn't support .net core.

mateusdemboski:
Any chance to be ported ou compatible in the future?
My project depend of this library... =(

ahmetalpbalkan:
@mateusdemboski well, you can try porting the .x509 package to .net core and send as a pull request, I'd happily package it for you.

jstarks:
I believe we will have a solution for this soon.

ahmetalpbalkan:
@swernli  so is this fixed with #50 merged?

swernli:
Unfortunately, no.  X509 auth implementation still depends on types that aren't present in DNXCore.  This will need to be fixed in a later PR.

jstarks:
Should be fixed now.

PR #63, opened 2016-05-05
Managed HTTP client

jstarks:
This adds a managed HTTP client that the ASP.NET team was working on and adds a bunch of necessary functionality (POST support with chunked encoding, client certificates, arbitrary (non-TCP) streams).

With this, we can support npipe for Docker on Windows, and we can (eventually) support UNIX domain sockets for Docker on Linux. We can also support TLS on CoreCLR with some minor build changes now.

Eventually this HTTP code should move outside the repo, but for now this is the best place for it.


Docker.DotNet.BasicAuth/BasicAuthHandler.cs
```diff
@@ -1,40 +1,25 @@
...
-#endif
+        private readonly MaybeSecureString _username;
+        private readonly MaybeSecureString _password;
 
-#if !netstandard
-        public BasicAuthHandler(SecureString username, SecureString password, HttpMessageHandler innerHandler) : base(innerHandler)
-#else
-        public BasicAuthHandler(string username, string password, HttpMessageHandler innerHandler) : base(innerHandler)
-#endif
+        public BasicAuthHandler(MaybeSecureString username, MaybeSecureString password, HttpMessageHandler innerHandler) : base(innerHandler)
```
ahmetalpbalkan: what is this `Maybe` magic

jstarks:
This will _almost_ fix #47. We just need to update project.json.

ahmetalpbalkan:
isn't there a nuget package for `Microsoft.Net.Http.Client`?

jstarks:
No... I believe they cut plans for this managed HTTP client in lieu of calling to winhttp and libcurl for Windows and Linux. The problem with that is that those don't support named pipe or unix sockets (well, libcurl might support unix sockets...), and they don't support connection hijacking, which we need for docker attach.

Docker.DotNet.BasicAuth/MaybeSecureString.cs
```diff
@@ -0,0 +1,74 @@
+using System;
+using System.Runtime.InteropServices;
+using System.Security;
+
+namespace Docker.DotNet.BasicAuth
+{
+    internal struct MaybeSecureString : IDisposable
+    {
+#if !netstandard
+        public SecureString Value;
```
jterry75: Make it an autoproperty instead of public member var. IE:
```C#
public SecureString Value { get; }
```

Docker.DotNet.BasicAuth/MaybeSecureString.cs
```diff
@@ -0,0 +1,74 @@
...
+                    secureStr.AppendChar(c);
+                }
+            }
+
+            Value = secureStr;
+        }
+
+        public MaybeSecureString(SecureString value)
+        {
+            Value = value;
```
jterry75: Is the assumption here that the callers validate that value was not null? Seems like it might be better to have validation here and you cant say
```C#
new MaybeSecureString(null);
```

Docker.DotNet.BasicAuth/MaybeSecureString.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        public void Dispose()
+        {
+            Value.Dispose();
```
jterry75: These seem to conflict. Value.Dispose() on null would throw but .Empty compares to null? How can we be in this case ever?

Docker.DotNet/AnonymousCredentials.cs
```diff
@@ -1,22 +1,11 @@
-﻿using System.Net.Http;
+using System.Net.Http;
 
 namespace Docker.DotNet
 {
     public class AnonymousCredentials : Credentials
     {
-        private readonly HttpClientHandler _handler;
-
         public AnonymousCredentials()
```
jterry75: You dont need it if there is no content to the constructor.

Docker.DotNet/AnonymousCredentials.cs
```diff
@@ -26,7 +15,11 @@ public override bool IsTlsCredentials()
 
         public override void Dispose()
         {
```
jterry75: We shouldnt override if you dont do anything special right?

Docker.DotNet.BasicAuth/MaybeSecureString.cs
jstarks: If we do this, then we can't pass the argument name into the exception. So it's a trade off. I don't have a strong feeling.

Docker.DotNet.BasicAuth/MaybeSecureString.cs
jstarks: Whoops. I also think I need to make sure we copy the thing since this guy basically takes ownership.

Docker.DotNet/AnonymousCredentials.cs
jstarks: I think Dispose is abstract

Docker.DotNet/DockerClient.cs
```diff
@@ -53,10 +56,75 @@ internal DockerClient(DockerClientConfiguration configuration, Version requested
...
+                    throw new Exception($"Unknown URL scheme {configuration.EndpointBaseUri.Scheme}");
+            }
+
+            _endpointBaseUri = uri;
 
-            _defaultTimeout = _client.Timeout;
+            ManagedHandler handler;
+            if (opener != null)
+            {
+                handler = new ManagedHandler(opener);
```
jterry75: If you can create one with a default constructor and one with an opener why not just handle the null case in the constructor? That way you can have one statement here for callers.

Docker.DotNet/Microsoft.Net.Http.Client/BufferedReadStream.cs
```diff
@@ -0,0 +1,242 @@
...
+
+        public override void SetLength(long value)
+        {
+            throw new NotSupportedException();
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            _disposed = true;
+            if (disposing)
```
jterry75: I think the pattern is better described
```C#
if (disposing && !_disposed)
{
}
```
Otherwise you could get a call with disposing == false but mark this object as disposed internally skipping the stream on the real dispose call.

Docker.DotNet/Microsoft.Net.Http.Client/BufferedReadStream.cs
```diff
@@ -0,0 +1,242 @@
...
+
+            // Drain buffer
+            if (_bufferCount > 0)
+            {
+                int toCopy = Math.Min(_bufferCount, count);
+                Buffer.BlockCopy(_buffer, _bufferOffset, buffer, offset, toCopy);
+                _bufferOffset += toCopy;
+                _bufferCount -= toCopy;
+                return toCopy;
+            }
```
jterry75: If toCopy < count you can do a small read at the same time after draining the buffer. Shouldn't you do that all at once?

Docker.DotNet/Microsoft.Net.Http.Client/BufferedReadStream.cs
```diff
@@ -0,0 +1,242 @@
...
+
+            return await _inner.ReadAsync(buffer, offset, count, cancellationToken);
+        }
+
+        private void EnsureBufferd()
+        {
```
jterry75: Buffered*

Docker.DotNet/Microsoft.Net.Http.Client/BufferedReadStream.cs
```diff
@@ -0,0 +1,242 @@
...
+                _bufferOffset = 0;
+                _bufferCount = _inner.Read(_buffer, _bufferOffset, _buffer.Length);
+                if (_bufferCount == 0)
+                {
+                    throw new IOException("Unexpected end of stream");
+                }
+            }
+        }
+
+        private async Task EnsureBufferdAsync(CancellationToken cancel)
```
jterry75: Buffered*

Docker.DotNet/Microsoft.Net.Http.Client/BufferedReadStream.cs
jterry75: Also for the sync async pairs a common pattern is to call

```C#
var t = this.EnsureBufferedAsync().ConfigureAwait(false);
t.Wait
```
instead of duplicating the code.

Docker.DotNet/Microsoft.Net.Http.Client/BufferedReadStream.cs
```diff
@@ -0,0 +1,242 @@
...
+                        foundCR = false;
+                    }
+                }
+            }
+            while (!foundCRLF);
+
+            return builder.ToString(0, builder.Length - 2); // Drop the CRLF
+        }
+
+        private void CheckDisposed()
```
jterry75: The intent here is more clear to me to call it "ThrowIfDisposed"

Docker.DotNet/Microsoft.Net.Http.Client/ChunkedReadStream.cs
```diff
@@ -0,0 +1,233 @@
...
+                {
+                    // End of chunk, read the terminator CRLF
+                    string trailerLine = _inner.ReadLine();
+                    System.Diagnostics.Debug.Assert(string.IsNullOrEmpty(trailerLine));
+                    _expectChunkHeader = true;
+                }
+
+                if (read == 0)
+                {
+                    Dispose();
```
jterry75: Is it common to have a call to Read end up with a disposed stream if it hits the end? Shouldnt this be left up to the client?

Docker.DotNet/Microsoft.Net.Http.Client/ChunkedReadStream.cs
```diff
@@ -0,0 +1,233 @@
...
+            catch (Exception)
+            {
+                Dispose();
+                throw;
+            }
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
```
jterry75: Same feedback about disposing && !_disposed

Docker.DotNet/Microsoft.Net.Http.Client/ChunkedReadStream.cs
```diff
@@ -0,0 +1,233 @@
...
+            {
+                // TODO: Sync drain with timeout if small number of bytes remaining?  This will let us re-use the connection.
+                _inner.Dispose();
+            }
+            _disposed = true;
+        }
+
+        private void CheckDisposed()
```
jterry75: Same feedback about ThrowIfDisposed

Docker.DotNet/Microsoft.Net.Http.Client/ChunkedWriteStream.cs
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        public override void SetLength(long value)
+        {
+            throw new NotImplementedException();
+        }
+
+        public override void Write(byte[] buffer, int offset, int count)
+        {
+            WriteAsync(buffer, offset, count, CancellationToken.None).Wait();
```
jterry75: I think you want ConfigureAwait(false); then task.Wait()

Docker.DotNet/Microsoft.Net.Http.Client/ChunkedWriteStream.cs
```diff
@@ -0,0 +1,83 @@
...
+        }
+
+        public override async Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+        {
+            if (count == 0)
+            {
+                return;
+            }
+                        
+            var chunkSize = Encoding.ASCII.GetBytes(Convert.ToString(count, 16) + "\r\n");
```
jterry75: Redefine these in this class for CR LF?

Docker.DotNet/Microsoft.Net.Http.Client/ContentLengthReadStream.cs
```diff
@@ -0,0 +1,158 @@
...
+            return read;
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                // TODO: Sync drain with timeout if small number of bytes remaining?  This will let us re-use the connection.
+                _inner.Dispose();
+            }
```
jterry75: You forgot to set _disposed = true

Docker.DotNet/Microsoft.Net.Http.Client/HttpConnection.cs
```diff
@@ -0,0 +1,169 @@
...
+
+                // Receive headers
+                List<string> responseLines = await ReadResponseLinesAsync(cancellationToken);
+                // Determine response type (Chunked, Content-Length, opaque, none...)
+                // Receive body
+                return CreateResponseMessage(responseLines);
+            }
+            catch (Exception ex)
+            {
+                Dispose(); // Any errors at this layer abort the connection.
```
jterry75: Im just not used to this pattern of disposing from within the call so pointing out again. Are we sure this is the right approach?

Docker.DotNet/Microsoft.Net.Http.Client/HttpConnection.cs
```diff
@@ -0,0 +1,169 @@
...
+        {
+            StringBuilder builder = new StringBuilder();
+            builder.Append(request.Method);
+            builder.Append(' ');
+            builder.Append(request.GetAddressLineProperty());
+            builder.Append(" HTTP/");
+            builder.Append(request.Version.ToString(2));
+            builder.Append(CRLF);
+
+            builder.Append(request.Headers.ToString());
```
jterry75: Since these are "static" should you use one call to Append($"{method}  {addressline} HTTP/{version}{CRLF}{headers}"

Docker.DotNet/Microsoft.Net.Http.Client/HttpConnection.cs
```diff
@@ -0,0 +1,169 @@
...
+                if (contentLength.HasValue)
+                {
+                    request.Content.Headers.ContentLength = contentLength.Value;
+                }
+
+                builder.Append(request.Content.Headers.ToString());
+                if (!contentLength.HasValue)
+                {
+                    // Add header for chunked mode.
+                    builder.Append("Transfer-Encoding: chunked\r\n");
```
jterry75: CRLF

Docker.DotNet/Microsoft.Net.Http.Client/HttpConnection.cs
```diff
@@ -0,0 +1,169 @@
...
+        public void Dispose()
+        {
+            Dispose(true);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                Transport.Dispose();
```
jterry75: set _disposed

Docker.DotNet/Microsoft.Net.Http.Client/HttpConnection.cs
jstarks: hmm... won't that build an extra string?

i guess I could use AppendFormat()

Docker.DotNet/Microsoft.Net.Http.Client/BufferedReadStream.cs
jstarks: The pattern here seems to be the canonical one, even if it's a little weird. The idea being that only the finalizer would ever call this with false, so who cares.

Docker.DotNet/Microsoft.Net.Http.Client/BufferedReadStream.cs
jstarks: No, I think that would be incorrect. _inner.Read would block if there is currently no data available. We must return the available data immediately.

Docker.DotNet/Microsoft.Net.Http.Client/BufferedReadStream.cs
jstarks: I'm going to leave the duplicated code for now.

Docker.DotNet/Microsoft.Net.Http.Client/HttpConnection.cs
jstarks: Fixed all this.

Docker.DotNet/Microsoft.Net.Http.Client/HttpConnection.cs
jstarks: I'm not a fan of this CRLF constant. I've removed it.

Docker.DotNet/Microsoft.Net.Http.Client/HttpConnection.cs
jstarks: Doesn't seem necessary since we can safely re-dispose the child objects.

Docker.DotNet.BasicAuth/MaybeSecureString.cs
jterry75: As long as we fix the null checks below then I dont care.

jstarks:
@jterry75 I've pushed a new commit that responds to most of your concerns. I also fixed ManagedHandler's proxy behavior to be more like HttpClientHandler's so that I can easily use Fiddler to debug this stuff.

Docker.DotNet/Microsoft.Net.Http.Client/ManagedHandler.cs
```diff
@@ -26,12 +26,24 @@ public ManagedHandler(StreamOpener opener = null)
...
+        public IWebProxy Proxy
         {
-            // TODO: Validate that only an absolute http address is specified. Path, query, and fragment are ignored
-            get; set;
+            get
+            {
+                if (_proxy == null)
+                {
+                    _proxy = WebRequest.GetSystemWebProxy();
+                }
```
jterry75: Are you doing this this way for a more lazy instantiation?

```C#
public IWebProxy Proxy { get; set; } = WebRequest.GetSystemWebProxy();
```

jterry75:
LGTM

Docker.DotNet/Microsoft.Net.Http.Client/ManagedHandler.cs
jstarks: That's right.

PR #64, opened 2016-05-05
WIP:Add AspNetCore services extension method AddDocker

eByte23:
Currently work in progress as I need to test further.

But this allows for simple setup of DI in aspnetcore app

### DIFF
--- a/Docker.DotNet.BasicAuth/project.json
+++ b/Docker.DotNet.BasicAuth/project.json
@@ -1,7 +1,8 @@
 {
   "version": "1.2.2",
   "dependencies": {
-    "Docker.DotNet": "2.124.0"
+    "Docker.DotNet": "2.125.0",
+    "Microsoft.Net.Http": "2.2.29"
   },
   "runtimes": {
     "win10-x64": {},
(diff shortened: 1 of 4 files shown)


Docker.DotNet.BasicAuth/project.json
```diff
@@ -1,7 +1,8 @@
 {
   "version": "1.2.2",
   "dependencies": {
-    "Docker.DotNet": "2.124.0"
+    "Docker.DotNet": "2.125.0",
```
jstarks: We are using the 124 to mean that this supports Docker API version 1.24.

jstarks:
@eByte23 can you help me understand whether this change should be in this module or if it could be separate? It seems like it could be implemented in your project without changing Docker.DotNet. I hesitate to add a dependency on Microsoft.Extensions.DependencyInjection.Abstractions just for a convenience method.

eByte23:
@jstarks sorry I did this super quick just it work at the time, I though that Microsoft.Extensions.DependencyInjection.Abstractions was already in the packages, but I now see it wasn't I totally understand your hesitation.
The only reason I suggested it on the first place is a lot library's do include an extension method like this for the easy of setup.

If you don't think it necessary please just close off the pr other wise I can fix up the pr.

jstarks:
@eByte23 Gotcha. I can see how this would be useful, but I think I'd like to avoid adding the extra dependency at this time. If there's a way we can include this without the extra dependency for users that aren't using dependency injection, I'm open to reconsidering.

Thank you for the contribution.

Issue #62, opened 2016-05-03
ImageBuildParameters.Tags is not getting serialized correctly


jstarks:
ERRO[6681] Handler for POST /v1.23/build returned error: Error parsing reference: "System.Collections.Generic.List`1[System.String]" is not a valid repository/tag

PR #66, opened 2016-05-05
Resolves: #62 - Correctly serializes list and dictionary in query

jterry75:

### DIFF
--- a/Docker.DotNet/EnumerableQueryStringConverter.cs
+++ b/Docker.DotNet/EnumerableQueryStringConverter.cs
@@ -1,0 +1,41 @@
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
+using Newtonsoft.Json;
+
+namespace Docker.DotNet
+{
+    /// <summary>
+    /// Handles serialization of objects like Lists, Arrays, etc.
+    /// </summary>
+    internal class EnumerableQueryStringConverter : IQueryStringConverter
+    {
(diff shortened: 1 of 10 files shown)


jterry75:
@jstarks @swernli - Looking for feedback here but this handles the issues that we are seeing for enumerables and maps.

Docker.DotNet/EnumerableQueryStringConverter.cs
```diff
@@ -0,0 +1,33 @@
...
+            return typeof (IEnumerable).IsAssignableFrom(t);
+        }
+
+        public string[] Convert(object o)
+        {
+            Debug.Assert(o != null);
+            Debug.Assert(o is IEnumerable);
+
+            var items = new List<string>();
+            foreach (var e in ((IEnumerable) o))
```
jstarks: Seems like this could be a lot simpler with linq

jterry75: Unfortunately LINQ only works on IEnumerable<T> and I dont see why we need a generic restriction on EnumerableQueryStringConverter but If you want I can make it EnumerableQueryStringConverter<T> and then on the attributes generate typeof(EnumerableQueryStringConverter<T>) that matches.

jstarks: ehh not worth it

jstarks:
LGTM

jterry75:
Verified that this now properly handles maps/lists.

PR #68, opened 2016-05-05
Adds the auto generator tool initial merge to Docker.DotNet

jterry75:


jterry75:
@jstarks There are still two outstanding PR's for specgen but they are small so we can either complete them there and merge across or I can just make the changes here inline with the PR's that they change for Docker.DotNet model code. Do you have a preference?

jterry75:
Resolves: https://github.com/ahmetalpbalkan/Docker.DotNet/issues/52

jterry75:
All outstanding specgen PR's have now been merged and this is ready.

jstarks:
Please squash the first 5 commits into one by me. We don't need commits with titles like "Stuff" :)

jstarks:
Can you add instructions on how to use this somewhere? Maybe it would be nice to set it up so that you just call go generate.

jterry75:
I was trying to do this just as a merge of the old repo then add the vendor/instructions. Would you prefer I do that all as one PR?

Issue #61, opened 2016-05-03
In Go null values for struct types are serialized/deserialized as the defaults for that type


jterry75:
We need to handle
```js
{
    Value: null
}
```
```Go
type Test struct {
    Value int
}
```
in C# the same way so that we end up with
```C#
class Test {
    int Value {get;set;}
}

// Value == 0
```
not an exception from Json.NET

jterry75:
References part of: https://github.com/Microsoft/Docker-PowerShell/issues/39

PR #69, opened 2016-05-06
Handle deserializing null values to defaults in C#

jterry75:
Resolves an issue where the Docker daemon might not send an actual value
that can be deserialized into the strongly typed C# model. In go values
like this are given the default value of the type itself but it does not
cause a failure in deserialization.

Resolves: #61


jterry75:
Have tested that this fixes the issue reported in Docker.PowerShell.

PR #74, opened 2016-05-07
Use NETSTANDARD1_3 instead of netstandard1_3

jstarks:
This appears to be a standard definition.


jterry75:
You just want them uppercase? Will send in a few mins.

PR #75, opened 2016-05-07
Support attaching to Docker container input/output

jstarks:
The key tricky part here is that we have to have a way to shut down the write path to notify Docker that stdin has closed. This is done with Socket.Shutdown() for sockets (TCP), and a zero-byte write for named pipe.

### DIFF
--- a/Docker.DotNet/DockerClient.cs
+++ b/Docker.DotNet/DockerClient.cs
@@ -174,6 +174,21 @@ namespace Docker.DotNet
             return new DockerApiStreamedResponse(response.StatusCode, body, response.Headers);
         }
 
+        internal async Task<WriteClosableStream> MakeRequestForHijackedStreamAsync(IEnumerable<ApiResponseErrorHandlingDelegate> errorHandlers, HttpMethod method, string path, IQueryString queryString, IDictionary<string, string> headers, IRequestContent data, CancellationToken cancellationToken)
+        {
+            HttpResponseMessage response = await MakeRequestInnerAsync(InfiniteTimeout, HttpCompletionOption.ResponseHeadersRead, method, path, queryString, headers, data, cancellationToken).ConfigureAwait(false);
+
+            HandleIfErrorResponse(response.StatusCode, null, errorHandlers);
+
+            var content = response.Content as HttpConnectionResponseContent;
+            if (content == null)
+            {
+                throw new NotSupportedException("message handler does not support hijacked streams");
+            }
+
+            return content.HijackStream();
+        }
+
         private Task<HttpResponseMessage> MakeRequestInnerAsync(TimeSpan? requestTimeout, HttpCompletionOption completionOption, HttpMethod method, string path, IQueryString queryString, IDictionary<string, string> headers, IRequestContent data, CancellationToken cancellationToken)
         {
             HttpRequestMessage request = PrepareRequest(method, path, queryString, headers, data);
--- a/Docker.DotNet/DockerClient.cs
+++ b/Docker.DotNet/DockerClient.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.IO;
+using System.IO.Pipes;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
@@ -81,13 +82,14 @@ namespace Docker.DotNet
                         // NamedPipeClientStream handles file not found by polling until the server arrives. Use a short
                         // timeout so that the user doesn't get stuck waiting for a dockerd instance that is not running.
                         int timeout = 100; // 100ms
-                        var stream = new System.IO.Pipes.NamedPipeClientStream(serverName, pipeName);
+                        var stream = new NamedPipeClientStream(serverName, pipeName);
+                        var dockerStream = new DockerPipeStream(stream);
 #if NET45
                         await Task.Run(() => stream.Connect(timeout), cancellationToken);
 #else
                         await stream.ConnectAsync(timeout, cancellationToken);
 #endif
-                        return stream;
+                        return dockerStream;
                     };
 
                     break;
@@ -113,7 +115,17 @@ namespace Docker.DotNet
 
             _endpointBaseUri = uri;
 
-            _client = new HttpClient(Configuration.Credentials.GetHandler(new ManagedHandler(opener)), true);
+            ManagedHandler handler;
+            if (opener == null)
+            {
+                handler = new ManagedHandler();
+            }
+            else
+            {
+                handler = new ManagedHandler(opener);
+            }
+
+            _client = new HttpClient(Configuration.Credentials.GetHandler(handler), true);
             _defaultTimeout = _client.Timeout;
             _client.Timeout = InfiniteTimeout;
         }
--- a/Docker.DotNet/DockerPipeStream.cs
+++ b/Docker.DotNet/DockerPipeStream.cs
@@ -1,0 +1,124 @@
+using System;
+using System.IO;
+using System.IO.Pipes;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Net.Http.Client;
+using Microsoft.Win32.SafeHandles;
+
+namespace Docker.DotNet
+{
+    internal class DockerPipeStream : WriteClosableStream
+    {
+        private PipeStream _stream;
+        private EventWaitHandle _event = new EventWaitHandle(false, EventResetMode.AutoReset);
+
+        public DockerPipeStream(PipeStream stream)
+        {
+            _stream = stream;
+        }
+
+        public override bool CanCloseWrite => true;
+
+        public override bool CanRead => true;
+
+        public override bool CanSeek => false;
+
+        public override bool CanWrite => true;
+
+        public override long Length
+        {
+            get { throw new NotImplementedException(); }
+        }
+
+        public override long Position
+        {
+            get { throw new NotImplementedException(); }
+
+            set { throw new NotImplementedException(); }
+        }
+
+        [DllImport("api-ms-win-core-file-l1-1-0.dll", SetLastError = true)]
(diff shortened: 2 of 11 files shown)


Docker.DotNet/DockerClient.cs
```diff
@@ -174,6 +176,21 @@ internal Task<DockerApiStreamedResponse> MakeRequestForStreamedResponseAsync(IEn
             return new DockerApiStreamedResponse(response.StatusCode, body, response.Headers);
         }
 
+        internal async Task<WriteClosableStream> MakeRequestForHijackedStreamAsync(IEnumerable<ApiResponseErrorHandlingDelegate> errorHandlers, HttpMethod method, string path, IQueryString queryString, IDictionary<string, string> headers, IRequestContent data, CancellationToken cancellationToken)
+        {
+            HttpResponseMessage response = await MakeRequestInnerAsync(InfiniteTimeout, HttpCompletionOption.ResponseHeadersRead, method, path, queryString, headers, data, cancellationToken).ConfigureAwait(false);
+
```
jterry75: Your in an "async" context. You don't need .ConfigureAwait(false) right?

Docker.DotNet/DockerPipeStream.cs
```diff
@@ -0,0 +1,124 @@
...
+using System.IO.Pipes;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Net.Http.Client;
+using Microsoft.Win32.SafeHandles;
+
+namespace Docker.DotNet
+{
+    internal class DockerPipeStream : WriteClosableStream
```
jterry75: Go has taken over your life.. `WriteClosableStream`. Should we post something like this to CoreFX as feature request and see what they say? I agree with its merits but would be interested to see what the core team thinks about how to bring this type of patterns to C#

Docker.DotNet/DockerPipeStream.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        public override bool CanCloseWrite => true;
+
+        public override bool CanRead => true;
+
+        public override bool CanSeek => false;
+
+        public override bool CanWrite => true;
+
+        public override long Length
```
jterry75: Not that I care but just FYI you can do these at autoprop's too.

```C#
public override long Length => throw new NotImp();
```

Docker.DotNet/DockerPipeStream.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        public override long Position
+        {
+            get { throw new NotImplementedException(); }
+
+            set { throw new NotImplementedException(); }
+        }
+
+        [DllImport("api-ms-win-core-file-l1-1-0.dll", SetLastError = true)]
+        private static extern int WriteFile(SafeHandle handle, IntPtr buffer, int numBytesToWrite, IntPtr numBytesWritten, ref NativeOverlapped overlapped);
```
jterry75: C# best practice is to add a file per assembly called NativeMethods.cs where all the PInvoke calls sit. It you dont like that then this is fine but it does make it super easy to see what is available.

jterry75: Also how does this work? Have you tried this? What is the "SafeHandle" deserialized as? I thought for CoreFX this was unsupported did I miss that it is now implemented?

Docker.DotNet/Endpoints/ContainerOperations.cs
```diff
@@ -28,6 +28,20 @@ internal ContainerOperations(DockerClient client)
             this.Client = client;
         }
 
+        public async Task<MultiplexedStream> AttachContainerAsync(string id, bool tty, ContainerAttachParameters parameters, CancellationToken cancellationToken)
```
jterry75: Most of the other methods have lightweight validation for things like `id`.

Docker.DotNet/Endpoints/ContainerOperations.cs
```diff
@@ -28,6 +28,20 @@ internal ContainerOperations(DockerClient client)
...
+        {
+            var path = $"containers/{id}/attach";
+            var queryParameters = new QueryString<ContainerAttachParameters>(parameters);
+            var stream = await this.Client.MakeRequestForHijackedStreamAsync(new[] {NoSuchContainerHandler}, HttpMethod.Post, path, queryParameters, null, null, cancellationToken).ConfigureAwait(false);
```
jterry75: Same. Do you need the `ConfigureAwait(false)`? Your in an async context.

Docker.DotNet/Microsoft.Net.Http.Client/ManagedHandler.cs
```diff
@@ -16,13 +16,23 @@ namespace Microsoft.Net.Http.Client
...
-            if (_opener == null)
+            _streamOpener = opener;
+            if (_streamOpener == null)
             {
-                _opener = TCPStreamOpener;
+                _socketOpener = TCPSocketOpener;
+            }
+        }
+
+        public ManagedHandler(SocketOpener opener = null)
```
jterry75: Seems odd to me. The compiler shouldn't know how to decode this style of default constructor.

IE:
```C#
new ManagedHandler() -> (SocketOpener default)
new ManagedHandler(null) -> err
new ManagedHandler((StreamOpener)null) -> (StreamOpener)
new ManagedHandler((SocketOpener)null) -> (SocketOpener)
```

Doesnt it make more sense to just say nulls aren't allowed and create an explicit default constructor that sets `TCPSocketOpener` to be more clear?

Docker.DotNet/Microsoft.Net.Http.Client/ManagedHandler.cs
```diff
@@ -141,7 +152,25 @@ private bool IsAllowedRedirectResponse(HttpRequestMessage request, HttpResponseM
...
 
             ProxyMode proxyMode = DetermineProxyModeAndAddressLine(request);
-            Stream transport = await ConnectAsync(request, cancellationToken);
+            Socket socket;
+            Stream transport;
+            try
+            {
+                if (_socketOpener != null)
+                {
+                    socket = await _socketOpener(request.GetConnectionHostProperty(), request.GetConnectionPortProperty().Value, cancellationToken).ConfigureAwait(false);
```
jterry75: The .Value call implies to me that its nullable. Do we always have a value? Do you want `GetValueOrDefault(80)`

Docker.DotNet/Microsoft.Net.Http.Client/WriteClosableStream.cs
```diff
@@ -0,0 +1,11 @@
+using System.IO;
+
+namespace Microsoft.Net.Http.Client
+{
+    public abstract class WriteClosableStream : Stream
```
jterry75: Why not interface?

Docker.DotNet/MultiplexedStream.cs
```diff
@@ -0,0 +1,161 @@
...
+                {
+                    case TargetStream.StandardIn:
+                    case TargetStream.StandardOut:
+                    case TargetStream.StandardError:
+                        _target = (TargetStream)_header[0];
+                        break;
+
+                    default:
+                        throw new IOException("unknown stream type");
+                }
```
jterry75:
```C#
if (!Enum.IsDefined(typeof(TargetStream), _header[0])
{
    throw IO;
}
```
_target = (TargetStream)_header[0];

Docker.DotNet/MultiplexedStream.cs
```diff
@@ -0,0 +1,161 @@
...
+            {
+                Count = read,
+                Target = _target
+            };
+        }
+
+        public async Task CopyFromAsync(Stream input, CancellationToken cancellationToken)
+        {
+            var buffer = new byte[bufferSize];
+            for (;;)
```
jterry75: while (true) ?

Docker.DotNet/MultiplexedStream.cs
```diff
@@ -0,0 +1,161 @@
...
+
+                Stream stream = null;
+                switch (result.Target)
+                {
+                    case TargetStream.StandardIn:
+                        stream = stdin;
+                        break;
+                    case TargetStream.StandardOut:
+                        stream = stdin;
+                        break;
```
jterry75: stdout?

Docker.DotNet/MultiplexedStream.cs
```diff
@@ -0,0 +1,161 @@
...
+                    case TargetStream.StandardError:
+                        stream = stdin;
+                        break;
```
jterry75: stderr?

Docker.DotNet/DockerPipeStream.cs
jstarks: This doesn't work since throw foo is a statement not an expression. It turns out the brace style fat arrow is not supported in this context, either.

Docker.DotNet/DockerPipeStream.cs
jstarks: This pattern works and seems to be used in corefx.

I'm going to hold off on moving this to NativeMethods until we understand how to support this in Linux.

Docker.DotNet/Microsoft.Net.Http.Client/ManagedHandler.cs
jstarks: I'm just following existing coding patterns here. I think it's fine.

Docker.DotNet/Microsoft.Net.Http.Client/WriteClosableStream.cs
jstarks: I need to extend Stream, which is already abstract.

Docker.DotNet/MultiplexedStream.cs
jstarks: fixed these

jstarks:
pushed updates

jterry75:
LGTM

PR #80, opened 2016-06-02
WIP: Fixes the generated model of Filters.

jterry75:
Fixes the filters query string paramter that was being incorrectly
generated based on the go client type as it is an un-exported type and
internally Docker Client-Types modify it before serialization.


jterry75:
@swernli - Give this a shot.

swernli:
Replaced by #81

PR #81, opened 2016-06-02
Fix generated model for Filters

swernli:
This updates and replaces #80.


swernli:
Reviewed in-person with @jterry75.  Merging.

PR #83, opened 2016-06-05
Handle System.PlatformNotSupported exception when checking to see if a proxy can be used.

Sumo-MBryant:
When running on DotNet Core on Linux, the ```System.PlatformNotSupported``` was firing when checking for proxy bypass. I could not see any way to set ```UseProxy``` to false from client code, and so propose that this exception is handled to enable functionality on platforms that don't support a proxy.

If a proxy is in fact required on a platform that doesn't support it, I expect the call to ```Proxy.GetProxy()``` to still fail as expected.


jterry75:
LGTM

mikecowgill:
This fixed the issue for me when trying to create a container running on Docker for Windows Beta. (hyper-v)

PR #85, opened 2016-06-30
Fixing Network creation.

swernli:
@jstarks @jterry75 
CreateNetworkAsync was incorrectly using GET instead of POST.


jstarks:
LGTM

PR #90, opened 2016-07-08
Add basic appveyor manifest

jstarks:
This just ensures that there are no build breaks. It doesn't do any testing.


swernli:
LGTM

PR #95, opened 2016-07-14
Update to the Docker v1.12.0-rc4 API

### LINKED ISSUE #82: GODEPS aren't honoring the vendored-in engin-api
This means we can't control the update cadence, but instead the generator picks up the most recent engine changes unintentionally.

### LINKED ISSUE #86: Regenerate types using engine-api that is vendored into docker 1.12
Now that docker 1.12 is close to release, we should regenerate types using the engine-api release that is vendored into docker. We should track that until 1.12/remote API 1.24 is completely final.

jterry75:
Resolves #82 
Resolves #86 

Updates the vendored directory to the same hashes as the Docker published
v1.12.0-rc4 API.

Regenerates the models based on this specific version and updates adds the
new ones.


Docker.DotNet/Models/CreateContainerParameters.Generated.cs
```diff
@@ -108,6 +114,12 @@ public CreateContainerParameters(Config Config)
         [DataMember(Name = "StopSignal", EmitDefaultValue = false)]
         public string StopSignal { get; set; }
 
+        [DataMember(Name = "StopTimeout", EmitDefaultValue = false)]
+        public long StopTimeout { get; set; }
```
jstarks: should this be a TimeSpan?

jterry75: Yes but unfortunately this is tricky. Lets make an issue for this for next release. We need to know what the destination target expects. IE: nanoseconds, seconds, etc. Because they are not standardized I will need to handle conversion more specifically through some additional attribute.

jstarks: OK let's hold off until 2.125.

jstarks:
LGTM

Issue #92, opened 2016-07-13
The Created property on a container should be a DateTime object


pcgeek86:
Right now, the `Created` property on a container is a `long`. This should be converted to a `System.DateTime`.

https://github.com/Microsoft/Docker.DotNet/blob/44da8e6df3e52729d42a43c945a2e92aace071f7/Docker.DotNet/Models/ContainerListResponse.Generated.cs

Cheers,
**Trevor Sullivan**
Microsoft MVP: Cloud & Data Center Management
https://trevorsullivan.net
https://twitter.com/pcgeek86

jstarks:
@jterry75 We're correctly performing the mapping from `int64` to `long` here, but maybe we should add a special case for this one to map to DateTime (presumably the int64 is Unix time). What do you think?

jterry75:
Easy enough to do. We already do this in many other places. Just a miss in the autogen code. For example "ImagesListResponse.Created" is a properly converted Unix time -> DateTime. PR on its way.

jterry75:
There are a few more.

pcgeek86:
Thanks @jterry75 @jstarks -- the current code works surprisingly well. Keep it up.

PR #96, opened 2016-07-14
Returns container created time as DateTime

jterry75:
Resolves: #92

Fixes a bug where container.Created was a long unix date time and converts
that to a C# System.DateTime object.


swernli:
LGTM

PR #104, opened 2016-07-21
code improvements

akoken:
* Unused methods removed
* Used nameof (...) instead of "..."
* Default constructors removed. Compiler already generates it by default.
* Pascal case for constant variables
* readonly where possible

### DIFF
--- a/Docker.DotNet/Endpoints/MiscellaneousOperations.cs
+++ b/Docker.DotNet/Endpoints/MiscellaneousOperations.cs
@@ -21,18 +21,16 @@ namespace Docker.DotNet
         {
             if (authConfig == null)
             {
-                throw new ArgumentNullException("authConfig");
+                throw new ArgumentNullException(nameof(authConfig));
             }
             var data = new JsonRequestContent<AuthConfigParameters>(authConfig, this.Client.JsonSerializer);
-
-            const string path = "auth";
-            return this.Client.MakeRequestAsync(this.Client.NoErrorHandlers, HttpMethod.Post, path, null, data);
+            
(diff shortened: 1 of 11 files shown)


Docker.DotNet/Endpoints/MiscellaneousOperations.cs
```diff
@@ -21,18 +21,18 @@ public Task AuthenticateAsync(AuthConfigParameters authConfig)
...
             if (authConfig == null)
             {
-                throw new ArgumentNullException("authConfig");
+                throw new ArgumentNullException(nameof(authConfig));
             }
             var data = new JsonRequestContent<AuthConfigParameters>(authConfig, this.Client.JsonSerializer);
 
-            const string path = "auth";
-            return this.Client.MakeRequestAsync(this.Client.NoErrorHandlers, HttpMethod.Post, path, null, data);
+            const string Path = "auth";
```
jterry75: Why declare the const at all? We could just put it in MakeRequestAsync and have the same "const" without an extra assignment correct?

jterry75:
@akoken - Thanks for the submission! I like the cleanup. Just left one comment that makes sense at all instances of the question.

Docker.DotNet/Endpoints/MiscellaneousOperations.cs
akoken: Absolutely you're right. I looked over the codebase, so didn't noticed. I will make the changes, Thanks @jterry75 !

jterry75:
LGTM. - Mind squashing your changes into 1?

Thanks again!

ahmetalpbalkan:
@jterry75 FYI you can squash yourself while merging and edit the commit message:

image

jterry75:
@ahmetalpbalkan - Fun fact. Thats awesome!

PR #106, opened 2016-07-25
Fixing the exec start parameters to be the full ExecConfig.

swernli:
@jterry75 @jstarks 
Promoting ExecConfig to a full generated object, where ContainerExecCreateParameters and ContainerExecStartParameters both have copy constructors from ExecConfig.  This will make it easier for callers who want to call both, and also allows for exec starts to pass the correct full set of params needed to perform an attach.

### DIFF
--- a/Docker.DotNet/Endpoints/ContainerOperations.cs
+++ b/Docker.DotNet/Endpoints/ContainerOperations.cs
@@ -369,6 +369,8 @@ namespace Docker.DotNet
             return this.Client.MakeRequestAsync(new[] { NoSuchContainerHandler }, HttpMethod.Post, path, queryParameters, null, null, cancellationToken);
         }
 
+        // StartContainerExecAsync will start the process specified by id in detach mode with no connected
+        // stdin, stdout, or stderr pipes.
         public Task StartContainerExecAsync(string id, CancellationToken cancellationToken)
         {
             if (string.IsNullOrEmpty(id))
@@ -385,7 +387,14 @@ namespace Docker.DotNet
             return this.Client.MakeRequestAsync(new[] { NoSuchContainerHandler }, HttpMethod.Post, path, null, data, null, cancellationToken);
         }
 
+        // StartAndAttachContainerExecAsync will start the process specified by id with stdin, stdout, stderr
+        // connected, and optionally using terminal emulation if tty is true.
         public async Task<MultiplexedStream> StartAndAttachContainerExecAsync(string id, bool tty, CancellationToken cancellationToken)
+        {
+            return await StartWithConfigContainerExecAsync(id, new ExecConfig() { AttachStdin = true, AttachStderr = true, AttachStdout = true, Tty = tty }, cancellationToken);
+        }
+
+        public async Task<MultiplexedStream> StartWithConfigContainerExecAsync(string id, ExecConfig eConfig, CancellationToken cancellationToken)
         {
             if (string.IsNullOrEmpty(id))
             {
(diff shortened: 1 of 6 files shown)


swernli:
DO NOT MERGE YET.  I found one additional update that I need to make...

swernli:
@jterry75 Ok, this is ready for review and merge now.  Testing the updated method in Docker-PowerShell and it works like a charm!

Docker.DotNet/Endpoints/ContainerOperations.cs
```diff
@@ -387,25 +387,26 @@ public Task StartContainerExecAsync(string id, CancellationToken cancellationTok
 
         public async Task<MultiplexedStream> StartAndAttachContainerExecAsync(string id, bool tty, CancellationToken cancellationToken)
         {
+            return await StartWithConfigContainerExecAsync(id, new ExecConfig(){Tty = tty}, cancellationToken);
```
jterry75: You dont need to `await` this just return the task.

Docker.DotNet/Endpoints/ContainerOperations.cs
```diff
@@ -387,25 +387,26 @@ public Task StartContainerExecAsync(string id, CancellationToken cancellationTok
 
         public async Task<MultiplexedStream> StartAndAttachContainerExecAsync(string id, bool tty, CancellationToken cancellationToken)
```
jterry75: We should probably add comments about why you would use one over the other now that we know why it hangs on input.

jterry75:
LGTM with the comments.

Docker.DotNet/Endpoints/ContainerOperations.cs
swernli: Since this method is async, don't I need to return MultiplexedStream not Task<MultiplexedStream>, otherwise I get this error:
<code>
C:\Coding\src\github.com\microsoft\Docker-PowerShell\src\Docker.DotNet\Docker.DotNet\Endpoints\ContainerOperations.cs(390,20): error CS4016: Since this is an async method, the return expression must be of type 'MultiplexedStream' rather than 'Task<MultiplexedStream>'
C:\Coding\src\github.com\microsoft\Docker-PowerShell\src\Docker.DotNet\Docker.DotNet\Endpoints\ContainerOperations.cs(388,46): warning CS1998: This async method lacks 'await' operators and will run synchronously. Consider using the 'await' operator to await non-blocking API calls, or 'await Task.Run(...)' to do CPU-bound work on a background thread.
</code>

Docker.DotNet/Endpoints/ContainerOperations.cs
swernli: Good point! I'll add that...

Docker.DotNet/Endpoints/ContainerOperations.cs
```diff
@@ -385,27 +387,30 @@ public Task StartContainerExecAsync(string id, CancellationToken cancellationTok
             return this.Client.MakeRequestAsync(new[] { NoSuchContainerHandler }, HttpMethod.Post, path, null, data, null, cancellationToken);
         }
 
+        // StartAndAttachContainerExecAsync will start the process specified by id with stdin, stdout, stderr
+        // connected, and optionally using terminal emulation if tty is true.
         public async Task<MultiplexedStream> StartAndAttachContainerExecAsync(string id, bool tty, CancellationToken cancellationToken)
         {
+            return await StartWithConfigContainerExecAsync(id, new ExecConfig() { AttachStdin = true, AttachStderr = true, AttachStdout = true, Tty = tty }, cancellationToken);
```
swernli: I went ahead and changed the approach here... Since the method is called `StartAndAttachContainerExecAsync` I updated it to actually connect to all the pipes.  Otherwise, if the exec process was created with `AttachStdin` set to true and then it was passed as false here, we end up with that nasty hang.  It didn't seem write that calling this method might hang if the exec was configured a certain way, so I expect that this will get the right behavior.

jstarks:
LGTM

PR #107, opened 2016-07-28
Fixing Pull Image and fixing Create Image to use Pull.

swernli:
@jterry75 @jstarks 
Create image was the only way to perform a pull, but it turns out that pull image is a superset of the create image functionality.  This change adds in pull and just has create call that method.
NOTE: It also cleans up some autogenerated types (getting rid of AuthConfigParameters which was a duplicate of AuthConfig), which will require clients to update the name of that class.


Docker.DotNet/Endpoints/ImageOperations.cs
```diff
@@ -51,7 +51,7 @@ internal ImageOperations(DockerClient client)
             }
 
             string path = string.Format(CultureInfo.InvariantCulture, "images/{0}/json", name);
-            DockerApiResponse response = await this.Client.MakeRequestAsync(new[] {NoSuchImageHandler}, HttpMethod.Get, path, null).ConfigureAwait(false);
+            DockerApiResponse response = await this.Client.MakeRequestAsync(new[] { NoSuchImageHandler }, HttpMethod.Get, path, null).ConfigureAwait(false);
```
jterry75: NIT: mind reverting the spaces changes?

Docker.DotNet/Endpoints/ImageOperations.cs
```diff
@@ -116,20 +116,24 @@ public Task TagImageAsync(string name, ImageTagParameters parameters)
             return this.Client.JsonSerializer.DeserializeObject<ImageSearchResponse[]>(response.Body);
         }
 
-        public Task<Stream> CreateImageAsync(ImagesCreateParameters parameters, AuthConfigParameters authConfig)
+        public Task<Stream> CreateImageAsync(ImagesCreateParameters parameters, AuthConfig authConfig)
```
jterry75: You need to keep the null check or you will change the result on `parameters.`

Docker.DotNet/Endpoints/ImageOperations.cs
swernli: Kind of... this is the spacing that gets inserted by VSCode when you use their auto-format for C# code, so anytime anyone edits this file and uses auto-formatting, it will just end up getting inserted again.

Docker.DotNet/Endpoints/ImageOperations.cs
swernli: Good point!

Docker.DotNet/Endpoints/ImageOperations.cs
```diff
@@ -141,28 +145,18 @@ public Task<Stream> PushImageAsync(string name, ImagePushParameters parameters,
...
-            {
-                throw new ArgumentNullException("authConfig");
-            }
-
             return new Dictionary<string, string>
             {
                 {
                     RegistryAuthHeaderKey,
-                    Convert.ToBase64String(Encoding.UTF8.GetBytes(this.Client.JsonSerializer.SerializeObject(authConfig)))
+                    Convert.ToBase64String(Encoding.UTF8.GetBytes(this.Client.JsonSerializer.SerializeObject(authConfig != null ? authConfig : new AuthConfig())))
```
jterry75: `authConfig ?? new AuthConfig();`

tools/specgen/resttag.go
```diff
@@ -6,13 +6,13 @@ import (
 )
 
 const (
-	Header = "header"
-	Body   = "body"
-	Query  = "query"
+	header = "header"
```
jterry75: Just for fun?

tools/specgen/specgen.go
```diff
@@ -38,7 +38,6 @@ type typeDef struct {
 var dockerTypesToReflect = []typeDef{
 
 	// POST /auth
-	{reflect.TypeOf(AuthConfigParameters{}), "AuthConfigParameters"},
```
jterry75: Its safe to leave this but say types.AuthConfig{} that way its explicitly included if they every change how the types are "inherited"

Docker.DotNet/Endpoints/ImageOperations.cs
jterry75: Odd. I would swear it was VSCode who removed them in the first place. I wonder if they recently made a change? Do what the auto formatter does.

tools/specgen/resttag.go
swernli: Makes golint happy.  The choices were comment them since they were public or make them private.  Since no one outside of this file used it, I made them private.

tools/specgen/specgen.go
swernli: Yup, makes sense.

jterry75:
LGTM

swernli:
Thanks! Merging!

PR #110, opened 2016-08-03
Updating image save/load methods.

swernli:
@jterry75 @jstarks 
This change expands on the MiscellaneousOperations methods for loading and saving images, and updates the ImageLoadParameters.Quiet to be required since DockerD defaults to true.

### DIFF
--- a/Docker.DotNet/Endpoints/IMiscellaneousOperations.cs
+++ b/Docker.DotNet/Endpoints/IMiscellaneousOperations.cs
@@ -21,7 +21,11 @@ namespace Docker.DotNet
 
         Task<Stream> GetImageAsTarballAsync(string name, CancellationToken cancellationToken);
 
+        Task<Stream> GetImagesAsTarballAsync(string[] names, CancellationToken cancellationToken);
+
         Task LoadImageFromTarballAsync(Stream stream, CancellationToken cancellationToken);
+
+        Task<Stream> LoadImageFromTarball(Stream stream, ImageLoadParameters parameters, CancellationToken cancellationToken);
 
         Task<Stream> BuildImageFromDockerfileAsync(Stream contents, ImageBuildParameters parameters, CancellationToken cancellationToken);
     }
--- a/Docker.DotNet/Endpoints/MiscellaneousOperations.cs
+++ b/Docker.DotNet/Endpoints/MiscellaneousOperations.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Globalization;
 using System.IO;
 using System.Net.Http;
 using System.Threading;
@@ -24,12 +23,12 @@ namespace Docker.DotNet
                 throw new ArgumentNullException(nameof(authConfig));
             }
             var data = new JsonRequestContent<AuthConfig>(authConfig, this.Client.JsonSerializer);
-            
+
             return this.Client.MakeRequestAsync(this.Client.NoErrorHandlers, HttpMethod.Post, "auth", null, data);
         }
 
         public async Task<VersionResponse> GetVersionAsync()
-        {            
+        {
             DockerApiResponse response = await this.Client.MakeRequestAsync(this.Client.NoErrorHandlers, HttpMethod.Get, "version", null).ConfigureAwait(false);
             return this.Client.JsonSerializer.DeserializeObject<VersionResponse>(response.Body);
         }
@@ -40,7 +39,7 @@ namespace Docker.DotNet
         }
 
         public async Task<SystemInfoResponse> GetSystemInfoAsync()
-        {            
+        {
             DockerApiResponse response = await this.Client.MakeRequestAsync(this.Client.NoErrorHandlers, HttpMethod.Get, "info", null).ConfigureAwait(false); ;
             return this.Client.JsonSerializer.DeserializeObject<SystemInfoResponse>(response.Body);
         }
@@ -51,7 +50,7 @@ namespace Docker.DotNet
             {
                 throw new ArgumentNullException(nameof(parameters));
             }
-           
+
             IQueryString queryParameters = new QueryString<ContainerEventsParameters>(parameters);
             return this.Client.MakeRequestForStreamAsync(this.Client.NoErrorHandlers, HttpMethod.Get, "events", queryParameters, null, cancellationToken);
         }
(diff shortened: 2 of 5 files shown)


Docker.DotNet/Endpoints/IMiscellaneousOperations.cs
```diff
@@ -21,8 +21,14 @@ public interface IMiscellaneousOperations
 
         Task<Stream> GetImageAsTarballAsync(string name, CancellationToken cancellationToken);
 
+        Task<Stream> GetMultipleImagesAsTarballAsync(string[] names, CancellationToken cancellationToken);
```
jterry75: Just `GetImagesAsTarballAsync` please. Imag'es' is already plural

Docker.DotNet/Endpoints/IMiscellaneousOperations.cs
```diff
@@ -21,8 +21,14 @@ public interface IMiscellaneousOperations
...
+
         Task LoadImageFromTarballAsync(Stream stream, CancellationToken cancellationToken);
 
+        Task<Stream> LoadImageFromTarballWithProgressAsync(Stream stream, CancellationToken cancellationToken);
```
jterry75: A side note. If we moved `Tar` from Docker PowerShell back to its own then we could have Docker.DotNet reference it too and we could take methods like this as `(TarStream stream, ...)`.

What do you think?

jterry75: Can you explain the naming here?

What is the difference in the return stream from `LoadImageFromTarballWithProgressAsync` and `LoadImageFromTarball`?

Docker.DotNet/Endpoints/MiscellaneousOperations.cs
```diff
@@ -64,32 +63,50 @@ public Task<Stream> MonitorEventsAsync(ContainerEventsParameters parameters, Can
...
+
             IQueryString queryParameters = new QueryString<CommitContainerChangesParameters>(parameters);
             DockerApiResponse response = await this.Client.MakeRequestAsync(this.Client.NoErrorHandlers, HttpMethod.Post, "commit", queryParameters, data).ConfigureAwait(false);
             return this.Client.JsonSerializer.DeserializeObject<CommitContainerChangesResponse>(response.Body);
         }
 
         public Task<Stream> GetImageAsTarballAsync(string name, CancellationToken cancellationToken)
         {
-            if (string.IsNullOrEmpty(name))
+            return GetMultipleImagesAsTarballAsync(new[] { name }, cancellationToken);
```
jterry75: Does this change the output tar? IE: Is the call to multiple "a tar of image tar's" or in the case its just 1 is it a single image tar?

Docker.DotNet/Endpoints/MiscellaneousOperations.cs
```diff
@@ -64,32 +63,50 @@ public Task<Stream> MonitorEventsAsync(ContainerEventsParameters parameters, Can
...
+        }
+
+        public Task<Stream> GetMultipleImagesAsTarballAsync(string[] names, CancellationToken cancellationToken)
+        {
+            string path = "images/get";
```
jterry75: Can we just hard code this in the path. We dont use it anywhere

Docker.DotNet/Endpoints/MiscellaneousOperations.cs
```diff
@@ -64,32 +63,50 @@ public Task<Stream> MonitorEventsAsync(ContainerEventsParameters parameters, Can
...
+            EnumerableQueryString queryString = null;
+
+            if (names.Length > 0)
```
jterry75: `names?.Length > 0`

Docker.DotNet/Endpoints/MiscellaneousOperations.cs
```diff
@@ -64,32 +63,50 @@ public Task<Stream> MonitorEventsAsync(ContainerEventsParameters parameters, Can
...
         {
+            return LoadImageFromTarball(stream, new ImageLoadParameters { Quiet = true }, cancellationToken);
+        }
+
+        public Task<Stream> LoadImageFromTarballWithProgressAsync(Stream stream, CancellationToken cancellationToken)
+        {
+            return LoadImageFromTarball(stream, new ImageLoadParameters { Quiet = false }, cancellationToken);
+        }
+
+        public Task<Stream> LoadImageFromTarball(Stream stream, ImageLoadParameters parameters, CancellationToken cancellationToken)
```
jterry75: I see the answer to my above question about "progress" now. Lets chat about this.

Docker.DotNet/Endpoints/IMiscellaneousOperations.cs
swernli: Done.

Docker.DotNet/Endpoints/MiscellaneousOperations.cs
swernli: A tar with one image and a tar with multiple images are the same internal format (just more entries, effectively), so we don't have to worry about this.

Docker.DotNet/Endpoints/MiscellaneousOperations.cs
swernli: Yeah, makes sense.

Docker.DotNet/Endpoints/MiscellaneousOperations.cs
swernli: Updated.

Docker.DotNet/Endpoints/IMiscellaneousOperations.cs
jstarks: I don't think we want to take/produce TarStream. Users will not always want to parse or produce a tar -- sometimes they will just want to pass in a FileStream, for example.

Docker.DotNet/Endpoints/MiscellaneousOperations.cs
```diff
@@ -64,32 +63,49 @@ public Task<Stream> MonitorEventsAsync(ContainerEventsParameters parameters, Can
...
+        {
             if (stream == null)
             {
                 throw new ArgumentNullException(nameof(stream));
             }
 
-            BinaryRequestContent data = new BinaryRequestContent(stream, "application/x-tar");            
-            return this.Client.MakeRequestAsync(new[] {ImageOperations.NoSuchImageHandler}, HttpMethod.Post, "images/load", null, data, null, cancellationToken);
+            IQueryString queryParameters = new QueryString<ImageLoadParameters>(parameters);
```
jterry75: how does this constructor handle null?

Docker.DotNet/Endpoints/MiscellaneousOperations.cs
jterry75: Spoke offline. Lets remove this and just use the overload with `ImageLoadParameters`

jterry75:
With the 2 fixes above LGTM.

PR #112, opened 2016-08-18
NetStandard1.6 and Unix Socket Support

swernli:
@jterry75 @jstarks 
This includes 3 commits, encapsulating the work for making the build work on netstarndard1.6 and adding support for unix sockets.  There is also a minor fix-up to the appveyor build version (should have been 2.124.2.{build}).


swernli:
Never mind... this still has the wrong nuget feed...

PR #113, opened 2016-08-18
Adding unix sockets and building for netstandard1.6

swernli:
@jterry75 @jstarks 
These changes prepare the build to work against netstandard1.6, including formatting updates for project.json.
They also introduce unix socket support for connections when running directly on Linux.

### DIFF
--- a/Docker.DotNet.BasicAuth/BasicAuthCredentials.cs
+++ b/Docker.DotNet.BasicAuth/BasicAuthCredentials.cs
@@ -1,5 +1,5 @@
 ﻿using System.Net.Http;
-#if !NETSTANDARD1_3
+#if (NET45 || NET46)
 using System.Security;
 #endif
 
@@ -17,7 +17,7 @@ namespace Docker.DotNet.BasicAuth
             return new BasicAuthHandler(_username, _password, innerHandler);
         }
 
-#if !NETSTANDARD1_3
+#if (NET45 || NET46)
         public BasicAuthCredentials(SecureString username, SecureString password, bool isTls = false)
             : this(new MaybeSecureString(username), new MaybeSecureString(password), isTls)
         {
--- a/Docker.DotNet/Microsoft.Net.Http.Client/UnixDomainSocketEndPoint.cs
+++ b/Docker.DotNet/Microsoft.Net.Http.Client/UnixDomainSocketEndPoint.cs
@@ -1,0 +1,100 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE-MIT.txt file for more information.
+
+#if NETSTANDARD1_6
+
+using System;
+using System.Runtime.InteropServices;
+using System.Diagnostics;
+using System.Text;
+using System.Net;
+using System.Net.Sockets;
+
+namespace Microsoft.Net.Http.Client
+{
+    internal sealed class UnixDomainSocketEndPoint : EndPoint
+    {
+        private const AddressFamily EndPointAddressFamily = AddressFamily.Unix;
+
+        private static readonly Encoding s_pathEncoding = Encoding.UTF8;
+
+        private static readonly int s_nativePathOffset = 2; // = offsetof(struct sockaddr_un, sun_path). It's the same on Linux and OSX
+
+        private static readonly int s_nativePathLength = 91; // sockaddr_un.sun_path at http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_un.h.html, -1 for terminator
+
+        private static readonly int s_nativeAddressSize = s_nativePathOffset + s_nativePathLength;
+
+        private readonly string _path;
+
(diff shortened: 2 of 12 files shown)


Docker.DotNet/Microsoft.Net.Http.Client/UnixDomainSocketEndPoint.cs
```diff
@@ -0,0 +1,110 @@
+#if NETSTANDARD1_6
```
jterry75: Exclude this file in the project.json for the other projects

swernli: I would... but VSCode doesn't like explicit excludes in the project.json.  The empty file isn't the end of the world...

swernli:
@jstarks Any input?

Docker.DotNet.BasicAuth/BasicAuthCredentials.cs
```diff
@@ -1,5 +1,5 @@
 ﻿using System.Net.Http;
-#if !NETSTANDARD1_3
+#if (!NETSTANDARD1_3 && !NETSTANDARD1_6)
```
jstarks: Since this is going to be the more common case in the future, should we instead invert this to NET45 || NET46 ?

swernli: Sure, I can make that update.

Docker.DotNet/Microsoft.Net.Http.Client/UnixDomainSocketEndPoint.cs
jstarks: Probably this file should have the .NET copyright statement at the top

Docker.DotNet/Microsoft.Net.Http.Client/UnixDomainSocketEndPoint.cs
```diff
@@ -0,0 +1,110 @@
...
+        private static readonly int s_nativePathLength;
+
+        private static readonly int s_nativeAddressSize;
+
+        private readonly string _path;
+
+        private readonly byte[] _encodedPath;
+
+        [DllImport("System.Native", EntryPoint = "SystemNative_GetDomainSocketSizes")]
+        internal static extern void GetDomainSocketSizes(out int pathOffset, out int pathSize, out int addressSize);
```
jstarks: I don't really understand why this works

jstarks: I have a feeling this is going to break with a newer version of .NET Core. We may want to follow up with the coreclr team to see if this is the recommended solution, if you haven't already done that. But this looks OK right now.

swernli: Magic.

swernli: Agreed on the likelihood of breakage if they change the way the native methods for Linux are exposed.  The right solution is for them to make their UnixDomainSocketEndPoint implementation public again instead of internal.  I can file an issue about this.

Docker.DotNet/Microsoft.Net.Http.Client/UnixDomainSocketEndPoint.cs
jstarks: OK, but now the copyright statement says read the LICENSE file, which is our license file, not the .NET one.

swernli: There is a license file in the same folder. Is that the right one to use?





On Thu, Aug 18, 2016 at 7:00 PM -0700, "John Starks" <notifications@github.com<mailto:notifications@github.com>> wrote:


In Docker.DotNet/Microsoft.Net.Http.Client/UnixDomainSocketEndPoint.cs<https://github.com/Microsoft/Docker.DotNet/pull/113#discussion_r75418438>:

> @@ -0,0 +1,110 @@
> +#if NETSTANDARD1_6


OK, but now the copyright statement says read the LICENSE file, which is our license file, not the .NET one.

-

swernli:
Ready to merge?

Docker.DotNet/Microsoft.Net.Http.Client/UnixDomainSocketEndPoint.cs
swernli: Ok, got some good clarification from @stephentoub over at https://github.com/dotnet/corefx/issues/10981, and I'll replace this call with hard coded values instead of using a call into a private extern wrapper.

swernli:
Ok, I think it's really ready this time...

jterry75:
LGTM

stefanprodan:
Hello,

Is version 2.124.2 going to be published on NuGet? I want to use the library with the unix socket connector.

PR #115, opened 2016-08-23
Minor code cleanup.

jterry75:
Refactors a few patterns of code design.
Removes redundant assignments.
Replaces string.Format with string interpolation.
Uses var or strong declaration.
Inline's path strings to avoid temporary assignment.


swernli:
LGTM

PR #122, opened 2016-09-09
Updates the start options type.

jterry75:
1. Fixes a bug in the ContainerStart operation where we were passing the
wrong value as a config option.


swernli:
Wow, nice catch!  LGTM.

Issue #124, opened 2016-10-04
QueryStringConverterInstanceFactory is not thread safe


Ecroyd:
This QueryStringConverterInstanceFactory file is not thread safe. `https://github.com/Microsoft/Docker.DotNet/blob/master/Docker.DotNet/QueryStringConverterInstanceFactory.cs`

**What version of Docker.DotNet?:**
Latest NuGet version

**Stacktrace:**
Object reference not set to an instance of an object.
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Collections.Generic.Dictionary`2.set_Item(TKey key, TValue value)
   at Docker.DotNet.QueryStringConverterInstanceFactory.GetConverterInstance(Type t)
   at Docker.DotNet.QueryString`1.GetKeyValuePairs()
   at Docker.DotNet.QueryString`1.GetQueryString()
   at Docker.DotNet.HttpUtility.BuildUri(Uri baseUri, Version requestedApiVersion, String path, IQueryString queryString)
   at Docker.DotNet.DockerClient.PrepareRequest(HttpMethod method, String path, IQueryString queryString, IDictionary`2 headers, IRequestContent data)
   at Docker.DotNet.DockerClient.MakeRequestInnerAsync(Nullable`1 requestTimeout, HttpCompletionOption completionOption, HttpMethod method, String path, IQueryString queryString, IDictionary`2 headers, IRequestContent data, CancellationToken cancellationToken)
   at Docker.DotNet.DockerClient.<MakeRequestAsync>d__40.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Docker.DotNet.ContainerOperations.<ListContainersAsync>d__6.MoveNext()

jterry75:
@Ecroyd - I just submitted a PR to fix this. Thanks for pointing it out. Want to take a look and try it out?

Ecroyd:
@jterry75 - Looks good, can not reproduce issue

PR #125, opened 2016-10-04
Make ConverterInstanceFactory threadsafe

jterry75:
Resolves: #124

Makes the QueryStringConverterInstanceFactory thread safe by using a
concurrent dictionary to store the actual instances.


swernli:
LGTM.  I'll wait to hear from @Ecroyd before merging.

PR #127, opened 2016-10-14
Add the 'Logs' parameter to the Attach operation's model.

enyim:
Added the "logs" parameter to the Attach command.

    logs – 1/True/true or 0/False/false, return logs. Default false.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/#/attach-to-a-container


enyim:
Meanwhile I managed to compile and run specgen.

Turns out it uses dependencies from way before 1.24, and running it generates 

- 13 new models
- updates 11 existing models

enyim:
Yeah, I'm going to create a new PR that tracks v1.24 of the API, I'm still looking why it is desynced.

Issue #128, opened 2016-10-14
Allow specifying "logs" parameter to the Container Attach operation


enyim:
**Output of `dotnet --info`:**

```
.NET Command Line Tools (1.0.0-preview2-003121)

Product Information:
 Version:            1.0.0-preview2-003121
 Commit SHA-1 hash:  1e9d529bc5

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.14393
 OS Platform: Windows
 RID:         win10-x64
```

**What version of Docker.DotNet?:**

```
2.124.1
```

**Steps to reproduce the issue:**
ContainerAttachParameters does not have a Logs property

**Additional information:**

This in relation of PR #127. 

I may be wrong but I did not find any mention of the Logs parameter in the engine-api sources, no matter which commit I checked. (https://github.com/docker/engine-api/blob/master/types/client.go)

What should be the proper course of action here? Update the generated file manually? Send a PR first to docker, then regen the client after it's accepted?

A faster option would be probably making QueryStringParameter (and related classes) public, so users could inherit from the Models and use their own when needed. 

This way we could use experimental/beta/newer API versions, with an older client (until it gets officially updated).

jterry75:
Ah interesting. This stamp was actually done on the v1.24-rc4 release of Docker as the release indicates. Now that the v1.24 release has been fully stamped I can update the vendor and regen they types. I will submit a PR for this today. Thanks!

jterry75:
This was incorrectly closed.

Issue #119, opened 2016-08-31
Add Swarm, Services, and Node models for Docker v1.12.0


jterry75:
In order to support the Docker Swarm scenarios we should have the model generator output Swarm, Services, and Node models so that we can add endpoints for those services as well.

jterry75:
@jstarks, @swernli - What do you guys think about this? Should we introduce another dll for Docker.Swarm or put this into Docker.DotNet?

stefanprodan:
I would use Docker.DotNet dll since Swarm mode info is part of [SystemInfoResponse](https://github.com/Microsoft/Docker.DotNet/blob/master/Docker.DotNet/Models/SystemInfoResponse.Generated.cs#L160). I would also rename [Info class](https://github.com/Microsoft/Docker.DotNet/blob/master/Docker.DotNet/Models/Info.Generated.cs) to SwarmInfo.

PR #129, opened 2016-10-17
Updates Models to release v1.12.0

jterry75:
Resolves: #128 - Adds the 'Logs' parameter to ContainerAttachParameters as
the docker cli didnt use this property even though the daemon supported
it.

Partially implements: #119 - Adds the swarm/services and node models but
doesnt add any endpoint behavior.


swernli:
LGTM

PR #133, opened 2016-10-20
Updated Docker documentation links

rversaw:
Looks like Docker changed their url for their api documentation, so I updated the link. The `v1.14` link also wasn't in use, so I dropped it as well.


swernli:
LGTM

Issue #140, opened 2017-01-03
Error in docker inspect when swapiness is not set


maxfridbe:
This may be an issue with Windows Containers in 
```
Docker Docker version 1.12.2-cs2-ws-beta, build 050b611
```
**Output of `dotnet --info`:**

```
using .net 4.5.2
```

**What version of Docker.DotNet?:**

```
  <package id="Docker.DotNet" version="2.124.3" targetFramework="net452" />
```

**Steps to reproduce the issue:**
1. create a container
2. inspect a container
3.

**What actually happened?:**
Exception in the async call:
```
{"Error converting value {null} to type 'System.Int64'. Path 'HostConfig.MemorySwappiness', line 1, position 2205."}
```

**What did you expect to happen?:**
if it is null then it should be null-able.

jterry75:
Interesting. Does the relevant `docker inspect` call from cmd show MemorySwappiness: null, or is it omitted?

PR #141, opened 2017-01-04
Fixes an issue with nullable primitive types.

jterry75:
Resolves: #140. When a `*primitive` in Go converts to C# the only possible
conversion is to call this a `primitive?` to maintain the nullablility.


swernli:
LGTM.

Issue #139, opened 2016-12-05
Docker.DotNet.Models.Config Volumes property should be flat string list


swernli:
Right now, Docker.DotNet.Models.Config Volumes property is generated as `Dictionary<string,object>` because the Go type is `map[string]struct{}`.  But it turns out that docker mostly treats this as an array of strings, since all it does is iterate over the keys and process the string values, ignoring the object.  As a convenience for users, we should provide either a wrapper or alternate typing with special marshal handling to make it clear that the property is really meant to be flat list of strings.

This ended up causing problems in Docker-Powershell, as seen in https://github.com/Microsoft/Docker-PowerShell/issues/167

swernli:
I have a suspicion there are more properties like this hiding among the many container creation properties.  We should see if we can identify those that fit this pattern.

PR #142, opened 2017-01-05
Fixes map[string]struct{} as dictionary

jterry75:
Resolves: #139 in that docker handles Config.Volumes very specifically.
Rather than being a map of string struct it actually only uses the keys
which serializes as a flat list of string.


swernli:
LGTM.

PR #146, opened 2017-01-11
Updates the Models to v1.13.0-rc5

jterry75:
Updates the code to reference the models based on the v1.13.0-rc5 release
of Docker.
Moves the previous old reference to engine-api to docker/api as the two
repos have been incorporated.


jterry75:
@swernli - I believe this is all the model fixes to support v1.13 as of the docker RC5 tag. I will do one fell swoop after they stamp the final version but I think this gets us 99% of the way there.

swernli:
From my perusal it looks good.  Are you ready to merge?

jterry75:
Yes thats fine. Its in an L2 branch so we have another shot before mainline.

Issue #108, opened 2016-07-28
Tasks that return status streams should be replaced with calls to IProgress<StreamStatus>


swernli:
Methods that return a `Task<Stream>` like `PullImageAsync` could instead return `Task<IProgress<Stream>>` to abstract away some of the details of how docker reports it's progress.  This should simplify patterns for callers who want to report that progress.

@jterry75 Can expand on what we are looking for here.

jterry75:
Instead of returning `Task<Stream>` we should return `Task`. Then add a parameter to the function that takes `IProgress<PullStatus> progress`. Each time that we receive the new "stream value" we deserialize it into the `PullStatus` and then `Report()` any waiter. The client looks like this:

```C#
var p = new Progress<Docker.DotNet.PullStatus>(status =>
{
// Do something with the status? Update a UI/PS/Service on the Sync Context of your choice.
});

await PullImageAsync(/*params*/, p);
```

jterry75:
I updated the tile which was very misleading. My response above is what we actually want. Not to change the return type to return progress but rather pass in an `IProgress` of the appropriate type so we can handle the status reporting to waiters.

galvesribeiro:
@jterry75 as mentioned on #143, Lets implement this issue.

I would like to align here the proper method signatures:

1. For methods that are today returning `Task<Stream>` we should do as you mentioned (here)[https://github.com/Microsoft/Docker.DotNet/issues/108#issuecomment-236969481], correct?
2. For methods that are long running monitoring, like the ones on `MISC` endpoint, which endlessly monitor events on the engine/swarm, I would suggest return an `IObservable<T>` where T is a structure with the general status update. That way, people don't need to await on that and instead can just subscribe. They can even use Rx to consume. As you may know, `IObservable<T>` is available on .Net so we don't need add any dependencies... What do you think?

When I got answer for those two I'll submit the PR.

Thanks

jterry75:
1. Agreed, I think anything that returns Task<stream> other than the image downloads can be done this way.
2. Correct that is the right pattern for Long Running Tasks, I would love to see that change!

galvesribeiro:
Perfect! Will do it tonight when get home. Thanks @jterry75

galvesribeiro:
@jterry75 it is done on #147.

I decided to not use `IObservable<T>` because it will be too big breaking change on the public APIs.

I think `IProgress<T>` is enough for that case. Specially that most of the time, the events are limited to an small number.

[...]

PR #147, opened 2017-01-18
Implement proper handling of Docker Streams

galvesribeiro:
*This PR implements #108*

Read and notify properly Docker event Streams in:

- [x] Image Operations
- [x] Container Logs
- [x] Container Stats 
- [x] `MISC` endpoint *Monitor Event*

This PR will be update with one commit for each part to make easier to review.


src/Docker.DotNet/Endpoints/ImageOperations.cs
```diff
@@ -109,28 +109,54 @@ public Task TagImageAsync(string name, ImageTagParameters parameters)
...
-            return this._client.MakeRequestForStreamAsync(this._client.NoErrorHandlers, HttpMethod.Post, "images/create", queryParameters, RegistryAuthHeaders(authConfig), null, CancellationToken.None);
+            var responseStream = await _client.MakeRequestForStreamAsync(_client.NoErrorHandlers, HttpMethod.Post, "images/create", queryParameters, RegistryAuthHeaders(authConfig), null, CancellationToken.None);
+            using (var reader = new StreamReader(responseStream))
+            {
+                while (responseStream.CanRead && !reader.EndOfStream)
+                {
+                    var line = await reader.ReadLineAsync();
+                    if (progress == null) continue;
+
+                    var @event = JObject.Parse(line);
```
jterry75: Why the complexity here? Cant JSON.NET just handle the deserialization in 1 object definition?

src/Docker.DotNet/Models/Events.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Docker.DotNet.Models
+{
+    public class EventMessage
```
jterry75: Ah. Add a line to specgen.go by /events
`{reflect.TypeOf(events.Message{}), "EventsMessage"},` so that this stays in sync with the docker types. This will also generate for you the Actor struct etc.

jterry75: Ok did some digging. I honestly cant tell from the docker code. I think we actually want jsonmessage.JsonMessage type here instead. @swernli does this seem right?

jterry75: This will reflect the JSONError and JSONProgress structs too which is the full message as far as I can tell.

src/Docker.DotNet/Endpoints/ContainerOperations.cs
```diff
@@ -238,7 +238,7 @@ public Task RemoveContainerAsync(string id, ContainerRemoveParameters parameters
             return this._client.MakeRequestAsync(new[] { NoSuchContainerHandler }, HttpMethod.Delete, $"containers/{id}", queryParameters);
         }
 
-        public Task<Stream> GetContainerLogsAsync(string id, ContainerLogsParameters parameters, CancellationToken cancellationToken)
+        public async Task GetContainerLogsAsync(string id, ContainerLogsParameters parameters, CancellationToken cancellationToken, IProgress<string> logReport = null)
```
galvesribeiro: Sure! Let me change that.

src/Docker.DotNet/Endpoints/ImageOperations.cs
galvesribeiro: The reason is that I can't find the object structure well defined for those events in go files... So I got the fields that matter and manually read them. If you can point me to the right structure I can make it deserialize direct to the object just like in the others.

src/Docker.DotNet/Models/Events.cs
galvesribeiro: I got it from here https://github.com/docker/docker/blob/master/api/types/events/events.go which are the only objects used by this method. Do you want submit a PR with the Go stuff changed? I don't have it here and would take a while to get it working...

galvesribeiro:
@jterry75 done. Now only thing missing is that Go thing...

galvesribeiro:
@jterry75 ?

src/Docker.DotNet/Endpoints/ContainerOperations.cs
```diff
@@ -254,6 +256,21 @@ public Task<Stream> GetContainerLogsAsync(string id, ContainerLogsParameters par
             return this._client.MakeRequestForStreamAsync(new[] { NoSuchContainerHandler }, HttpMethod.Get, $"containers/{id}/logs", queryParameters, null, cancellationToken);
         }
 
+        public async Task GetContainerLogsAsync(string id, ContainerLogsParameters parameters, CancellationToken cancellationToken, IProgress<string> progress = null)
```
jterry75: Does this compile? I would assume that for this transfer period what we will need to do is to make the progress a non default parameter. Then we will have 2 physical overloads and then when I merge/remove the [Obsolete] version I will make it optional again to keep the same set of overloads.

src/Docker.DotNet/Endpoints/ContainerOperations.cs
```diff
@@ -427,5 +445,24 @@ public Task<Stream> GetContainerStatsAsync(string id, ContainerStatsParameters p
...
+            var responseStream = await GetContainerStatsAsync(id, parameters, cancellationToken);
+
+            using (var reader = new StreamReader(responseStream))
+            {
+                while (responseStream.CanRead && !reader.EndOfStream)
+                {
+                    var line = await reader.ReadLineAsync();
+                    if (progress == null) continue;
+
+                    var report = JsonConvert.DeserializeObject<ContainerStatsResponse>(line);
```
jterry75: Can you use the `this._client.JsonSerializer.DeserializeObject` it makes it a whole lot easier for writing tests to not hard reference JsonConvert

jterry75: And then you dont need the using at the top

src/Docker.DotNet/Endpoints/IContainerOperations.cs
```diff
@@ -40,6 +41,8 @@ public interface IContainerOperations
 
         Task<Stream> GetContainerLogsAsync(string id, ContainerLogsParameters parameters, CancellationToken cancellationToken);
```
jterry75: Ya now I think I mislead you. We only expose the Interfaces externally. I think you need to make the [Obsolete] tags here instead. My bad

src/Docker.DotNet/Endpoints/ImageOperations.cs
jterry75: Go ahead and leave this for now. I will remove it in the 1.13 branch that my PR eludes to because the type wasnt yet defined.

src/Docker.DotNet/Endpoints/MiscellaneousOperations.cs
```diff
@@ -4,6 +4,7 @@
 using System.Threading;
 using System.Threading.Tasks;
 using Docker.DotNet.Models;
+using Newtonsoft.Json;
```
jterry75: Same. If you remove the JsonConvert reference and use `this._client.JsonSerializer` we can remove this too

src/Docker.DotNet/Models/Events.cs
jterry75: Added this in the other PR. Rebase on it and you can remove.

src/Docker.DotNet/Models/ImageOperationProgress.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Docker.DotNet.Models
+{
+    public struct ImageOperationProgress
```
jterry75: Keep for now. I will remove later in the 1.13 branch

galvesribeiro:
@jterry75 done :)

PR #149, opened 2017-01-19
Implement /events return type.

jterry75:
For the v1.12.0 release the return type of /events was the
events.Message type in go. This adds that type to the reflection api's
to generate it for the return streams.

### DIFF
--- a/src/Docker.DotNet/Models/Actor.Generated.cs
+++ b/src/Docker.DotNet/Models/Actor.Generated.cs
@@ -1,0 +1,15 @@
+using System.Collections.Generic;
+using System.Runtime.Serialization;
+
+namespace Docker.DotNet.Models
+{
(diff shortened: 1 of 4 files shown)


jterry75:
@galvesribeiro - This is partially what you need. Unfortunately the merger from the api-types repo to the docker/docker/api happened for the 1.13 timeframe. So here is what I can do in "master". You can rebase your changes on this and then we can pull them up. I will merge to the v1.13 branch and do the correct fix which is to make the JSONMessage, JSONError and JSONProgress types generically across all event based API's.

galvesribeiro:
@jterry75 1.13 was already released today :)

jterry75:
@swernli - I am merging for now. Let me know if you have any comments and I can fix retroactively since this is add-only

PR #150, opened 2017-01-19
Replaces ContainerEvents with jsonmessage

jterry75:
The docker API now has a better /pkg/jsonmessage that handles all tasked based results for progress/errors/results. Replacing the EventsMessage with this for the v1.13.0 API.


jterry75:
@galvesribeiro - And the update for the v1.13 branch as promised.

PR #151, opened 2017-01-19
Fixes IProgress<stream> for v1.13

### LINKED ISSUE #108: Tasks that return status streams should be replaced with calls to IProgress<StreamStatus>

jterry75:
Resolves #108.

Full implements the IProgress<Stream> changes for the v1.13 api change
using JSONMessage JSONProgress and JSONError


jterry75:
@galvesribeiro - After this is merged. I will update to the final v1.13 that was released from docker (if any changes). Then merge that branch back to mainline master! Almost there! Thanks for all the help

galvesribeiro:
Thank you all for being so helpful :)

After everything get done, I'll have a look at set the tests up. :)

pbielinski:
IMO obsolete annotation was set on incorrect method GetContainerLogsAsync with IProgress

jterry75:
@pbielinski - You are 100% correct!

PR #152, opened 2017-01-19
Update models to release v1.13.0

jterry75:
This updates the vendor and models to the released final version v1.13.0
from docker.


jterry75:
There were only vendor changes so merging this as it doesnt affect any C# code.

PR #170, opened 2017-02-10
Multiple issues fixed, see details

jasoncouture:
CancellationToken support (#166)
Timeout configuration support (#168)
~~Fixed project files for VS2017 RC as noted here: https://aka.ms/sdkimplicititems (#169)~~

There are a few changed API calls, any interface method where CancellationToken was present, but was not the last parameter was reordered to put CancellationToken last, and make it optional. The majority of methods are unaffected, but some such as ```Task GetContainerStatsAsync(string id, ContainerStatsParameters parameters, IProgress<JSONMessage> progress, CancellationToken cancellationToken = default(CancellationToken));``` were

I also made some parameters optional where there was default behavior for the call, I did not do this everywhere as I was not explcitly looking for it. The one example I changed was: LoadImageFromTarball, parameters was optional in the method, but not the parameters. So it was made optional.

PR #179, opened 2017-02-27
Fixes stack overflow recursive call.

jterry75:
Replaces the make request method overloads with a proper forwarding chain
that exposes all of the possible overloads for MakeRequest,
MakeRequestForStream, MakeRequestForHijackedResponse

Resolves: #170


jterry75:
@pauldotknopf - This should fix your issue. Want to give me a quick review?

PR #192, opened 2017-03-22
Add import image

jonso4:
Add import image, both from an existing stream and from a url


jonso4:
I will add "save" in another pull request :)

jterry75:
I'm so confused now. I just re-read all the docs here and it looks like we are doing this wrong in three places now haha. What you implemented is actually the full `CreateImageAsync` impl and we should probably replace `CreateImageAsync` with this version that accepts an optional stream in the body. We should remove `PullImage*` because that doesn't actually make sense from the REST API sense and just have `CreateImage*` variants. Then we should use the methods you just provided `ImportImage` to actually do the `"images/load"` REST endpoint. What do you think?

jonso4:
Just read the docs again...you're definitely right. And along with that, we need to add the progress messages to the `CreateImageAsync` signature. So let's iron out what the interface will look like: we'll remove both `PullImageAsync` methods, and the existing `CreateImageAsync` methods. In their place, we'll have:

`Task CreateImageAsync(ImagesImportParameters parameters, AuthConfig authConfig, IProgress<JSONMessage> progress, CancellationToken cancellationToken = default(CancellationToken));`
`Task CreateImageAsync(ImagesImportParameters parameters, Stream imageStream, AuthConfig authConfig, IProgress<JSONMessage> progress, CancellationToken cancellationToken = default(CancellationToken));`
`Task LoadImageAsync(ImageLoadParameters parameters, Stream imagesStream, IProgress<JSONMessage> progress, CancellationToken cancellationToken = default(CancellationToken)`

Thoughts?

jterry75:
I think your right on all accounts and those look good. For the `Load` `IProgess` will cover that and given that the `parameters` are only `Quiet` we could just use the progress to determine that. What about:

```C#
Task LoadImageAsync(Stream imagesStream, CancellationToken);
Task LoadImageAsync(Stream imagesStream, IProgress<JSONMessage> progress, CancellationToken);
```

We know that the 1'st uses the images/load?quiet=true because there is no progress reporting possible.

jonso4:
Sounds good! I'll update this PR.

jonso4:
I decided to go with just one `Load` method in case people still wanted to get the final response message.

src/Docker.DotNet/Endpoints/ImageOperations.cs
```diff
@@ -108,47 +110,59 @@ public Task TagImageAsync(string name, ImageTagParameters parameters, Cancellati
...
         }
 
-        public Task<Stream> CreateImageAsync(ImagesCreateParameters parameters, AuthConfig authConfig)
+        public Task CreateImageAsync(ImagesCreateParameters parameters, AuthConfig authConfig, IProgress<JSONMessage> progress, CancellationToken cancellationToken)
         {
-            if (parameters == null)
-            {
-                throw new ArgumentNullException(nameof(parameters));
-            }
+            return CreateImageAsync(parameters, authConfig, progress, cancellationToken);
```
jterry75: I think you meant:

```C#
return CreateImageAsync(paramters, null, authConfig, progress, cancellationToken);
```

src/Docker.DotNet/Endpoints/ImageOperations.cs
```diff
@@ -164,13 +178,17 @@ public Task<Stream> PushImageAsync(string name, ImagePushParameters parameters,
             return this._client.MakeRequestForStreamAsync(this._client.NoErrorHandlers, HttpMethod.Post, $"images/{name}/push", queryParameters, null, RegistryAuthHeaders(authConfig), CancellationToken.None);
         }
 
-        public Task PushImageAsync(string name, ImagePushParameters parameters, AuthConfig authConfig, IProgress<JSONMessage> progress, CancellationToken cancellationToken = default(CancellationToken))
+        private Task<Stream> LoadImageAsyncPrivate(ImageLoadParameters parameters, Stream imageStream, IProgress<JSONMessage> progress, CancellationToken cancellationToken)
```
jterry75: You dont actually need these private methods. Just do the validation in the public one and pass this._client.MakeRequestForStreamAsync to the Monitor call.

src/Docker.DotNet/Endpoints/ImageOperations.cs
```diff
@@ -164,13 +178,17 @@ public Task<Stream> PushImageAsync(string name, ImagePushParameters parameters,
...
-                PushImageAsync(name, parameters, authConfig),
-                this._client,
-                cancellationToken,
-                progress);
+            if (parameters == null)
+            {
+                throw new ArgumentNullException(nameof(parameters));
+            }
+
+            BinaryRequestContent content = new BinaryRequestContent(imageStream, TarContentType);
```
jterry75: Youll want to validate imageStream too here

jterry75:
I like it!!! Thanks for doing this. Have you had a chance to do a quick smoke test before I merge the changes?

jonso4:
I tested it yesterday - won't get a chance to today, unfortunately. I can test it Monday, then let ya know when it's all set?

jterry75:
Yea that sounds great. Thanks again

jonso4:
Quick smoke test seems good! I don't have a way of testing out pulling from docker hub / auth right now, but we haven't changed too much here besides the interface, so I think it should be alright. What about you?

jterry75:
Looks good. Merging.

Issue #98, opened 2016-07-14
Some of the timeout style properties should be TimeSpan


jterry75:
For example:

`Config.StopTimeout`
`HealthConfig.Interval`
`HealthConfig.Timeout`
etc.

This will require us to implement custom JsonConverters for TimeSpan but because sometimes the result is in Seconds and in others its NanoSeconds we likely need to introduce an attribute that describes the destination conversion expected.

jasoncouture:
Just a comment on this rather than opening a new issue:
the underlying HttpClient timeout should be exposed, there are numerous situations where it would be highly helpful to be able to change the default timeout of 100 seconds.

client.RequestTimeout or configuration.RequestTimeout are a few suggestions.

jterry75:
@jasoncouture - Actually these arent the same type of "timeout" issue. Can you open a new issue for this so we can track it independently?

jasoncouture:
@jterry75 Sorry for the late response, Sure thing.

orthros:
So I've got a rough draft of this working, though it is untested.

specgen is now outputting certain types to ```TimeSpan?``` and you can specify on a per-property basis custom Attributes to decorate the property with.

On the json serialization of ```TimeSpan``` attributes: would it be better to decorate the Property with the "standard" ```[JsonConvert(Type converterType)]``` style, or to put in a ```IContractResolver``` and custom ```Attributes``` on the Properties to denote what should be serialized to/from seconds/nanoseconds? 

I have it working both ways, but prefer the custom attributes path to help keep the consuming code clean.

jasoncouture:
@orthros, 
Just my $0.02, I think the attribute is the better choice, because if we ever needed to change the contract resolver for something else it would be a headache. (I'm not a member of this project, it's just my opinion.)

orthros:
@jasoncouture thanks for the feedback! In my PR (#205 ), I wound up going with a single attribute that you can decorate with a Seconds or Nanoseconds enumeration like so

```csharp
[DataMember(Name = "Timeout", EmitDefaultValue = false)]
[TimeSpanSerialization(SerializationTarget.Nanoseconds)]
public TimeSpan Timeout { get; set; }
```
which then gets intercepted in the ContractResolver.

I wanted to avoid coupling the contract too closely with the serializer, and in his initial description @jterry75 mentioned introducing an Attribute, which I interpreted to be "custom" attribute 😄 

I do see your point though; messing with the ContractResolver can be opaque at times. 

[...]

PR #205, opened 2017-05-01
TimeSpan Feature

orthros:
For Issue #98 Added support for ```TimeSpan```
This required a change both in specgen as well as the JsonConvert

specgen now has a second map[] that will map a CSType to CSAttribute, which are applied in ```reflectTypeMembers```

Added an attribute ```TimeSpanSerializationAttribute``` to decorate Properties to denote if they should be serialized in Seconds or Nanoseconds

Here is a sample of the  generated code
```csharp
    [DataContract]
    public class HealthConfig // (container.HealthConfig)
    {
        [DataMember(Name = "Test", EmitDefaultValue = false)]
        public IList<string> Test { get; set; }

        [DataMember(Name = "Interval", EmitDefaultValue = false)]
        [TimeSpanSerialization(SerializationTarget.Nanoseconds)]
        public TimeSpan Interval { get; set; }

        [DataMember(Name = "Timeout", EmitDefaultValue = false)]
        [TimeSpanSerialization(SerializationTarget.Nanoseconds)]
        public TimeSpan Timeout { get; set; }

        [DataMember(Name = "Retries", EmitDefaultValue = false)]
        public long Retries { get; set; }
    }
```

Wrote a very simple test program and got passing results.

### DIFF
--- a/src/Docker.DotNet/JsonSerializer.cs
+++ b/src/Docker.DotNet/JsonSerializer.cs
@@ -14,7 +14,8 @@ namespace Docker.DotNet
         {
             JsonConvert.DefaultSettings = () => new JsonSerializerSettings
             {
-                NullValueHandling = NullValueHandling.Ignore
+                NullValueHandling = NullValueHandling.Ignore,
+                ContractResolver = new TimeSpanContractResolver()
             };
         }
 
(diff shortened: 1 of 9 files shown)


jterry75:
I personally prefer #207 solution. Do you have a preference?

orthros:
I prefer #207 as well. I made two separate PRs to foster discussion on what the better approach might be (especially with some of the comments in #98 )

jterry75:
Closing since we approved the other one #207.

PR #207, opened 2017-05-25
Timespan for Docker classes

orthros:
For Issue #98 Added support for TimeSpan

specgen now has a second map[] that will map a ```CSType``` to ```CSAttribute```, which are applied in ```reflectTypeMembers```

Also ```calcUsings``` will also take into account the namespaces of attributes attached to the Properties in a class

On the generated classes, now decorate the properties of type ```Timespan``` with the ```[JsonConverter]``` attribute 

This is an alternative to #205, which required custom attributes and a change to the ContractResolver for the Json serializer.

Here is a sample of the generated code
```csharp
[DataContract]
public class HealthConfig // (container.HealthConfig)
{
    [DataMember(Name = "Test", EmitDefaultValue = false)]
    public IList<string> Test { get; set; }

    [DataMember(Name = "Interval", EmitDefaultValue = false)]
    [JsonConverter(typeof(TimeSpanNanosecondsConverter))]
    public TimeSpan Interval { get; set; }

    [DataMember(Name = "Timeout", EmitDefaultValue = false)]
    [JsonConverter(typeof(TimeSpanSecondsConverter))]
    public TimeSpan Timeout { get; set; }

    [DataMember(Name = "Retries", EmitDefaultValue = false)]
    public long Retries { get; set; }
}
```


jterry75:
Cool this looks great! @swernli you good with this?

PR #209, opened 2017-06-06
Fix CreateImageAsync HTTP method

faulpeltz:
fix images/create route to always use POST instead of GET for requests with FromImage


swernli:
LGTM

PR #210, opened 2017-06-09
Added example of named pipes URI to connect to local Docker for Windows

elialgranti:
Took me a long time to figure out how to connect to the local daemon without exposing it on port 2375 (should have started with the code).
This should really be documented as it is a common use case.


swernli:
This is a great addition, thanks!

PR #211, opened 2017-06-09
Fixing swarm inspect type and adding Nodes types

swernli:
This change updates the swarm inspect type to match the type returned by docker (swarm.Swarm) and
adds the types for /nodes/ endpoint operations.


swernli:
@galvesribeiro I think this fixes the types for swarm that you mentioned, and it also introduces the node types that you could then use to implement the /nodes/ endpoint operations.  Take a look and let me know if this lines up with your expectations, and we can get this merged to master so you can get unblocked!

galvesribeiro:
It's all good @swernli. Thanks for the prompt reply! I'll implement the Nodes operations and submit a PR once I got it done.

Thank you!

PR #216, opened 2017-06-16
Dropped suport for .NET 4.5, Add .NET 4.5.2

alexgwolff:
### Dropped suport for .NET  4.5 

- There is no official support by Microsoft [Source](https://www.microsoft.com/net/download/framework)
- There is no installer for .NET 4.5 SDK
- .NET Core SDK can't build the project without .NET 4.5

### Add global.json
-  There's no ways the build project if .NET Core SDK 2 Preview is installer on dev machine


ikkentim:
In my opinion, it's rather pointless to drop support for .NET 4.5, unless it blocks the project from moving forwards/using a useful library/API

galvesribeiro:
@alexgwolff what is the failure you have if you build with .net core SDK 2.0 preview? Maybe we can workaround.

For the record, I'm all up to move everything to a `netstandard` library. However, there are *big* customers of this library which are still on `net45` and that can't be migrated.

Like @ikkentim mentioned, we need to know why to upgrade the minimum framework level to get it done.

PR #217, opened 2017-06-18
fix image operations create http method

liqdfire:
The call to "images/create" was set as an http get unless you passed a stream source. This results in a 404 when trying to pull an image from a registry.

repro snippet:

```c#
using System;
using Docker.DotNet;
using Docker.DotNet.Models;

namespace Docker_API_Testing
{
  class Program
  {
    static void Main(string[] args)
    {
      var client = new DockerClientConfiguration(new Uri("tcp://127.0.0.1:2375")).CreateClient();
      IProgress<JSONMessage> report = new Progress<JSONMessage>(msg =>
      {
        Console.WriteLine(msg.Status);
      });

      var parms = new ImagesCreateParameters { FromImage = "alpine", Tag = "3.5"};

      client.Images.CreateImageAsync(parms, null, report).GetAwaiter().GetResult();

    }
  }
}

```

### DIFF
--- a/src/Docker.DotNet/Endpoints/ImageOperations.cs
+++ b/src/Docker.DotNet/Endpoints/ImageOperations.cs
@@ -122,19 +122,17 @@ namespace Docker.DotNet
                 throw new ArgumentNullException(nameof(parameters));
             }
 
-            HttpMethod httpMethod = HttpMethod.Get;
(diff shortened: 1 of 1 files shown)


liqdfire:
Did not see the other PR, #209,  before I submitted this, closing as dupe PR.

PR #218, opened 2017-06-18
Fixed timeout for a couple of container operations

liqdfire:
The calls to container [wait, stop, restart] were using a timeout of -1 ticks, which causes an immediate cancellation request on the linked token. 

The infinite tmeout static is defined as -1 ms
```c#
private static readonly TimeSpan s_InfiniteTimeout = TimeSpan.FromMilliseconds(Timeout.Infinite);
```

Changed the calls in ContainerOperations to use 
```c#
TimeSpan.FromMilliseconds(Timeout.Infinite)
```


swernli:
LGTM

PR #221, opened 2017-06-27
ContainerListResponse should be used instead of ContainerResponse

sebagomez:
I'm new to DockerDotNet but I guess ContainerResponse is an old class


swernli:
Thanks for the fix!

Issue #158, opened 2017-01-20
Implement the Swarm Tasks API surface.


jterry75:
On the `ISwarmOperations` interface we need to add the `/tasks/*` API surface.

galvesribeiro:
@jterry75 I saw someone implemented Swarm operations... Can we have at least a pre-release nuget? I'm doing some work for MSR/Orleans teams internally and I need to use Swarm.

Thanks

swernli:
Sorry for the lack of response... Justin is out of office for a while, so we will be able to circle back on this once he gets back.

galvesribeiro:
Hey @swernli thanks for get back. Will ping you internally.

jterry75:
Working on this. Will submit for PR today.

PR #228, opened 2017-07-12
Implements the Swarm Task's API

jterry75:
Resolves: #158

### DIFF
--- a/src/Docker.DotNet/DockerClient.cs
+++ b/src/Docker.DotNet/DockerClient.cs
@@ -35,6 +35,8 @@ namespace Docker.DotNet
 
         public ISwarmOperations Swarm { get; }
 
+        public ITasksOperations Tasks { get; }
+
         public IVolumeOperations Volumes { get; }
 
         public TimeSpan DefaultTimeout { get; set; }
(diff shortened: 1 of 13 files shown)


galvesribeiro:
@jterry75 I only have 1 question... Why [TaskResponse](https://github.com/Microsoft/Docker.DotNet/pull/228/files#diff-b1361d745fe87c1826eb153cffcc4e2b) isn't being generated by the codegen from Go source like all other models? (just curious)

jterry75:
I was looking at the API and absent minded just wrote the model. I have updated the PR with the automated changes. (And good thing, the documentation is missing some the models allow!)

galvesribeiro:
Yeah... There are some stuff not well documented there.

I wish we could move to use Swagger and drop the codegen. Would make our life easier.

jterry75:
> I wish we could move to use Swagger and drop the codegen. Would make our life easier.

I haven't given up here. Thinking I might just write a yaml to roslyn transform and then we absolutely can do this for both the models and the api. But will take some time.

galvesribeiro:
> Thinking I might just write a yaml to roslyn transform

Yes, that was my initial plan. I thought on use https://www.nuget.org/packages/YamlDotNet/4.2.1

It would make us have 2 things:

1. Drop the Go codegen and use the yaml spec
2. Have the codegen be able to run either in compile or runtime mode -> Runtime mode will allow live update to new APIs. We have something like that in Orleans for Serializers.

jterry75:
Exactly. Although in practice the API isn't likely to change very often and simply being able to take 1 stamp per release and rerun the codegen should be sufficient. Its just that we today have to do a lot of manual work to figure out what the API even is from code rater than from the yaml.

PR #231, opened 2017-07-20
Dispose http client

Kagamine:


src/Docker.DotNet/DockerClient.cs
```diff
@@ -388,8 +388,12 @@ internal HttpRequestMessage PrepareRequest(HttpMethod method, string path, IQuer
         public void Dispose()
         {
             Configuration.Dispose();
+            if (_client != null)
+            {
+                _client.Dispose();
+            }
```
bgrainger: I suggest the use of the C# 6 null-coalescing operator: `_client?.Dispose();`.

bgrainger: Actually, the `if (_client != null)` appears superfluous because `_client` is assigned in the constructor and never cleared.

galvesribeiro:
I'm not sure why the error you stated is happening. Keep in mind that by design, call `Dispose()` on HttpClient will **not**  release the socket/port right away. It take some time to unmanaged code (i.e. WinHTTP) to release it. HttpClient was designed to be shared/reused. If you are disposing on every request, you are doing it wrong.

galvesribeiro:
Can we close this PR? This issue was fixed in another PR and already pushed to latest release.

PR #234, opened 2017-07-21
fix #233 , close socket after SendAsync

### LINKED ISSUE #233: Local ports run out after repeatedly call docker api
Refer to #232 , I changed DockerClient to singleton but it still cause local ports run out.

Code:

``` csharp
public async Task StartKeepaliveLoop()
{
    var client = dockerClientConfiguration.CreateClient();
    while (true)
    {
        await client.Miscellaneous.GetVersionAsync();
        await Task.Delay(1000);
    }
}
```

close_waits

Also I can reproduce this problem on my local machine:

303248153:
This PR fixed problem described in #233 ,

The reason of this problem:

First, create a docket client will create a ManageHandler with no ctor arguments:
https://github.com/Microsoft/Docker.DotNet/blob/9a740b51305180dcc87e0e9bc976e185a8e4e8ac/src/Docker.DotNet/DockerClient.cs#L108

Then, ManageHandler will use TCPSocketOpenerAsync by default:
https://github.com/Microsoft/Docker.DotNet/blob/master/src/Docker.DotNet/Microsoft.Net.Http.Client/ManagedHandler.cs#L23

TCPSocketOpenerAsync will open a new Socket for every request, but ProcessRequestAsync wont Dispose them
https://github.com/Microsoft/Docker.DotNet/blob/master/src/Docker.DotNet/Microsoft.Net.Http.Client/ManagedHandler.cs#L188

I already test this fix with my local environment and confirm it can prevent socket resource leak.
Maybe you guys want implement socket reuse but for now please take this solution and release a new version...

### DIFF
--- a/src/Docker.DotNet/DockerClient.cs
+++ b/src/Docker.DotNet/DockerClient.cs
@@ -200,12 +200,14 @@ namespace Docker.DotNet
             CancellationToken token)
         {
             var response = await PrivateMakeRequestAsync(timeout, HttpCompletionOption.ResponseContentRead, method, path, queryString, headers, body, token).ConfigureAwait(false);
+            using (response)
+            {
+                var responseBody = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
 
-            var responseBody = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+                HandleIfErrorResponse(response.StatusCode, responseBody, errorHandlers);
 
-            HandleIfErrorResponse(response.StatusCode, responseBody, errorHandlers);
-
-            return new DockerApiResponse(response.StatusCode, responseBody);
+                return new DockerApiResponse(response.StatusCode, responseBody);
+            }
         }
 
         internal Task<Stream> MakeRequestForStreamAsync(
--- a/src/Docker.DotNet/Microsoft.Net.Http.Client/ManagedHandler.cs
+++ b/src/Docker.DotNet/Microsoft.Net.Http.Client/ManagedHandler.cs
@@ -149,8 +149,8 @@ namespace Microsoft.Net.Http.Client
             request.Headers.ConnectionClose = true; // TODO: Connection re-use is not supported.
 
             ProxyMode proxyMode = DetermineProxyModeAndAddressLine(request);
-            Socket socket;
-            Stream transport;
+            Socket socket = null;
+            Stream transport = null;
             try
             {
                 if (_socketOpener != null)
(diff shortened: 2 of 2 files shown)


galvesribeiro:
Oh! I forget we are using a different handler!

Thanks for the PR, I'll have a look at it in the morning.

@jterry75 do you have anything against this fix?

303248153:
Wait, I found another problem, some request need read the stream after ProcessRequestAsync  returned, this fix will cause an ObjectDisposedException, so don't merge it yet.

303248153:
The old fix doesn't work for stream response so I did a new fix, it contains
- change new NetworkStream(socket) to new NetworkStream(socket, true)
- dispose HttpResponseMessage after read it as String

And the code I used to confirm there no resource leak with this fix:

``` csharp
static void DockerLoop()
{
	var cert = new CertificateCredentials(new X509Certificate2(CertPath, CertPassword));
	var configuration = new DockerClientConfiguration(new Uri("https://docker-1:2376"), cert);
	var client = configuration.CreateClient();
	while (true)
	{
		var resultA = client.System.GetVersionAsync().Result;
		var resultB = client.Containers.GetArchiveFromContainerAsync(
			"90bdd9842f5e",
			new GetArchiveFromContainerParameters() { Path = "/workdir/abc.txt" },
			false).Result;
		resultB.Stream.Dispose();
		Thread.Sleep(1);
	}
}
```

src/Docker.DotNet/DockerClient.cs
```diff
@@ -200,12 +200,14 @@ internal DockerClient(DockerClientConfiguration configuration, Version requested
             CancellationToken token)
         {
             var response = await PrivateMakeRequestAsync(timeout, HttpCompletionOption.ResponseContentRead, method, path, queryString, headers, body, token).ConfigureAwait(false);
```
jterry75: We should likely have any calls thought the client to `PrivateMakeRequestAsync` dispose of the `HttpResponseMessage` correct?

src/Docker.DotNet/Microsoft.Net.Http.Client/ManagedHandler.cs
```diff
@@ -156,7 +156,7 @@ private bool IsAllowedRedirectResponse(HttpRequestMessage request, HttpResponseM
                 if (_socketOpener != null)
```
jterry75: Its possible its just my state of tired but what are the changes in this file? They look like space only?

ikkentim: It indeed is, Add ?w=1 to ignore whitespace (so you can see the difference) https://github.com/Microsoft/Docker.DotNet/pull/234/files/c2456a2a599b914ac7b4047ef9652b575220bf95?w=1

src/Docker.DotNet/DockerClient.cs
jterry75: Ah nvm. The other 3 are all streams and we cant close them in the method body.

galvesribeiro:
Thanks @303248153 for the fix! :)

PR #235, opened 2017-07-21
Adds SystemOperations tests.

### LINKED ISSUE #176: Add unit tests
I would like to add unit tests as described in readme.md
I propose NUnit. If that is okay, I will fork and get started as soon as I can.

jterry75:
Adds a null check to the progress for event monitoring.
Adds a cancellation check when an events stream is cancelled.

Partially Resolves #176


jterry75:
@galvesribeiro - I am going to add at least a few tests per service so that we can have some coverage. (Long overdue). Things like the connection and the images we use I will clean up into some helpers at the end. For now I'm going to hard code it.

Issue #240, opened 2017-07-26
UpdateServiceAsync : Invalid URL and request body


aemery:
**Output of `dotnet --info`:**

```
.NET Command Line Tools (2.0.0-preview2-006497)

Product Information:
 Version:            2.0.0-preview2-006497
 Commit SHA-1 hash:  06a2093335

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.15063
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.0.0-preview2-006497\

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.0-preview2-25407-01
  Build    : 40c565230930ead58a50719c0ec799df77bddee9
```

**What version of Docker.DotNet?:**

```
Built this morning following the d68e0ba commité
```

SwarmOperations.UpdateServiceAsync should use the endpoint "services/{id}/update" not "services/{id}" and add the body to the MakeRequestAsync call. It is currently defined but unused.

PR #241, opened 2017-07-26
Fixes a bug in UpdateServiceAsync.

jterry75:
Resolves: #240


galvesribeiro:
Was about to fix it. My bad... Bad copy and paste when implemented that method.

jterry75:
I have for sure done that.... 👍

Issue #248, opened 2017-08-02
Shouldn't this version be 3.125?


sebagomez:
AssemblyInfo.cs file still says 2.124, shouldn't it be 2.125 or 3.125?

PR #249, opened 2017-08-03
Fixes the assembly version numbers

jterry75:
Resolves: #248


jterry75:
@galvesribeiro - I think this is what we want but dont merge until I can check AppVeyor artifacts. This is correct for local builds for sure though.

galvesribeiro:
I think it must be `$($env:appveyor_build_version)` instead of `{build}`

jterry75:
Well that is really unfortunate. I dont really want to have the build number in two places. But if you dont set the `version:` tag in the appveyor.yaml the version is defaulted to 1.0.0.{build}. So the nuget feed shows it as say 1.0.0.81 while the assembly version is 3.125.0.81. I put that tag back but now we always have to remember to rev the version in the csproj and the appveyor.yaml. Do you have a better idea?

galvesribeiro:
I don't personally like AppVeyor. We use Jenkins in Orleans which I don't like as well, but at least it works in that aspect. Will think about it. For now lets move as is.

Issue #254, opened 2017-09-18
Support /images/prune


aemery:
**Output of `dotnet --info`:**

```
.NET Command Line Tools (2.0.0)

Product Information:
 Version:            2.0.0
 Commit SHA-1 hash:  cdcd1928c9

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.15063
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.0.0\

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.0
  Build    : e8b8861ac7faf042c87a5c2f9f2d04c98b69f28d
```

**What version of Docker.DotNet?:**

```
Github's tip
```

The API offers a /images/prune to delete all unused images. /volumes/prune and /networks/prune are already available in here, but not the one for images.

Would it be possible to add it as well? Thanks!

By the way /volumes/prune maps to IVolumeOperations.PruneAsync and /networks/prune maps to INetworkOperations.DeleteUnusedNetworksAsync, if you ever want to normalize the nomenclature.

PR #255, opened 2017-09-18
Resolves issues with /*/prune methods

jterry75:
Resolves: #254


jterry75:
@aemery - Want to give this change a shot. It should solve all your issues I think.

aemery:
@jterry75  Works great. Again, thank you for your quick feedback.

Issue #258, opened 2017-09-22
GetContainerStatsAsync throws DockerApiException if container is not found


lmolkova:
**Output of `dotnet --info`:**

```
.NET Command Line Tools (2.0.0)

Product Information:
 Version:            2.0.0
 Commit SHA-1 hash:  cdcd1928c9

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.15063
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.0.0\

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.0
  Build    : e8b8861ac7faf042c87a5c2f9f2d04c98b69f28d

```

**What version of Docker.DotNet?:**
```
3.125.0
```

**Steps to reproduce the issue:**
Call `GetContainerStatsAsync` with non-existing container
```C#
            var statsParameters = new ContainerStatsParameters { Stream = true };
            dockerClient.Containers.GetContainerStatsAsync("12345", statsParameters, CancellationToken.None).Wait();
```
**What actually happened?:**
`DockerApiException` is thrown

```
Unhandled Exception: System.AggregateException: One or more errors occurred. (Docker API responded with status code=NotFound, response=) ---> Docker.DotNet.DockerApiException: Docker API responded with status code=NotFound, response=
   at Docker.DotNet.DockerClient.HandleIfErrorResponse(HttpStatusCode statusCode, String responseBody, IEnumerable`1 handlers)
   at Docker.DotNet.DockerClient.<MakeRequestForStreamAsync>d__50.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Docker.DotNet.Models.StreamUtil.<MonitorStreamForMessagesAsync>d__1`1.MoveNext()
```

**What did you expect to happen?:**
DockerContainerNotFoundException

**Additional information:**

Looking at the [code](https://github.com/Microsoft/Docker.DotNet/blob/master/src/Docker.DotNet/Endpoints/ContainerOperations.cs#L437)

``` C#
        public Task<Stream> GetContainerStatsAsync(string id, ContainerStatsParameters parameters, CancellationToken cancellationToken)
        {
            if (string.IsNullOrEmpty(id))
[...]

PR #259, opened 2017-09-25
Fixes an issue in GetContainerStatsAsync

jterry75:
Adds the NoSuchContainerHandler to GetConatinerStatsAsync that was missing.
Reorders the ContainerOperations and IContainerOperations to match the API spec.
Adds the appropriate HTTP URI to the documents for each function.

Resolves: #258


jterry75:
@lmolkova - This should fix your issue. Sorry for all the reordering as there are really no changes other than adding the handler.

Issue #264, opened 2017-10-16
Stream Length property unavailable from MultiplexedStream


aemery:
**What version of Docker.DotNet?:**

```
Tip
```

EDIT 2: tl;dr StartContainerExecAsync that returns the body of the API response and with a parameter to set detach to false is all I need. Sorry for the double edit, should have dug a little more before opening the issue.

This is more of an "ease of use" request.

The MultiplexedStream class, used to get the output of /exec/{id}/start, does not expose the Length property of the internal _stream object. When calling ReadOutputAsync, an arbitrary number must be given as count. Passing something like multiplexedStream.Length would make more sense IMO.

An overloaded method with only buffer and cancellationToken as parameters could also be exposed by MultiplexedStream. In there, offset and count could be set internally to 0 and _stream.Length since this is the typical use case.

In any way, this is not a deal breaker, feel free to give me your opinion on this issue.

EDIT: Digging around in the SDK, I realized that simply having access to the body of the response to "/exec/{id}/start" as a string would cover my use case since I only need to get the output of a PowerShell command. i.e.

jterry75:
I'm not sure I fully understand. In the "attached" case how would you know the length? I am almost certain that even if we exposed the length it would throw because the actual stream doesn't know the answer. The concept here is that you can get any length of output. So it might take multiple buffered reads to get the full output.

Are you saying that you would like one full "Read to end and assign it to a string" method so that its just easier than having to do the reads yourself?

aemery:
Yep, exactly. I need a simple way to call a command on a Docker and get the complete output.

PR #265, opened 2017-10-16
Implements MultiplexedStream ReadToEndAsync

jterry75:
Resolves: #264

### DIFF
--- a/src/Docker.DotNet/Docker.DotNet.csproj
+++ b/src/Docker.DotNet/Docker.DotNet.csproj
@@ -19,6 +19,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="System.ValueTuple" Version="4.4.0" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.6' ">
     <PackageReference Include="System.Buffers" Version="4.3.0" />
--- a/src/Docker.DotNet/MultiplexedStream.cs
+++ b/src/Docker.DotNet/MultiplexedStream.cs
@@ -1,5 +1,6 @@
 using System;
 using System.IO;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Net.Http.Client;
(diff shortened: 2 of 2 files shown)


jterry75:
@aemery - I have not tried this but let me know if this is what you were looking for.

aemery:
This is indeed what I'm looking for. The output is not what I would expect though.

Command: powershell Get-Item 'HKLM:/SOFTWARE/ODBC/ODBCINST.INI/ODBC Drivers' | Select -ExpandProperty Property

Expected output: PostgreSQL ANSI(x64)\r\nPostgreSQL Unicode(x64)\r\nSQL Server
What I'm getting: "PostgreSQL ANSI(x64)PostgreSQL ANSI(x64)\r\nPostgreSQL Unicod" or "PostgreSQL ANSI(x64)\r\nPostgreSQL ANSI(x64)\r\nPPoPostgreSQLPo" (output is not constant when running the same code multiple times)

Using the REST Api directly (using Postman), I get the correct expected output.

aemery:
Sleeping between StartWithConfigContainerExecAsync and ReadOutputToEndAsync fixes it. Is there a way to ensure that the output is fully available before reading it? I'm basically trying to get the same behavior as calling docker exec or /exec/id/start in attached mode.

jterry75:
Ok looked at this harder. I understand the issue your hitting with this code fix. I can provide a better impl if you are ok with this assumption. Otherwise You really should use `CopyOutputToAsync` yourself because you know how it was configured.

`ReadOutToEndAsync` will really need to return a` Task<(string, string)>` where the results are `stdout, stderr`. Because on a true multiplexed stream we cannot assume that the output is just `stdout` or just `stderr` and there is no way to know how to make a single string from the output if using both.

galvesribeiro:
@jterry75  did you finished the implementation so can I have a look on it?

jterry75:
I did but I was hoping that @aemery would tell me if it solves his situation before we merge.

aemery:
Sorry for the delay, having the two output separated is ok for me. If you can provide a fix for the issue I mentionned above, I'm good with receiving a Task<(string, string)>.

jterry75:
@aemery - I actually pushed another fix already that should work and give you the full output. Want to give it a shot?

aemery:
Just tested your last fix. Works great.

PR #268, opened 2017-10-24
Adding support for .NETStandard2.0

jterry75:


jterry75:
I'm going to merge this. I don't think there is much risk given it builds just fine after adding netstandard2.0

galvesribeiro:
@jterry75 Sorry, I was out. Yeah it looks good!

jterry75:
No worries. I will be sending in the set of models / api changes one by one over the next few days to get us to v1.33 from Docker so look out for them. Thanks again for your help.

Issue #262, opened 2017-09-26
Swarm Configs support (API 1.30+)


mot256:
We need access to the [Swarm Configs](https://docs.docker.com/engine/api/v1.31/#tag/Config) calls on the Docker Remote API. As far as we can tell, the API supports those calls from version 1.30

Are there any plans to support API 1.30+ any time soon?

jterry75:
I am actively working on this. POR is that I am 1 by 1 going to make the set of commits from v1.25 where we are now to v1.33 where Docker is. This will include the models and api changes where applicable.

m-barthelemy:
Hi,
Is this still something that's going to be implemented at some point?

Crozin:
Any chance to update the lib to reflect changes made in newer API versions (current: 1.39). Mostly `Swarm.Configs` and `/update` actions for existing resources.

Crozin:
If we could at least update the autogenerated models then I could prepare PR with required changes.

jterry75:
Docker pretty fundamentally changed the config model. At this point I think it would actually be easier to just change the models from autogen to static. And then make the updates based on the swagger.yaml. @galvesribeiro - Should we just remove specgen and we can just update the models for each release of the Docker Remote API? I could do this work pretty easily I think.

jterry75:
Sorry that was unclear what I said. I mean Docker changed how the vendor model of their configs works for Docker EE/CE so it makes it hard to autogen

m-barthelemy:
Hi, any progress around stopping using autogenerated models? Or about implementing Swarm Configs?

On my end I tried to help and implement a minor feature quite some time ago but gave up when I saw this autogenerated models thing that I didn't know how to use/update.

jterry75:
#432. This is now v1.41

PR #269, opened 2017-10-27
Fixes an issue with specgen

jterry75:
Resolves a few issues with specgen that were causing it to improperly
disambiguate types with name collisions. This can happen for a few reasons
but namely happens when to types swarm.IPAM and network.IPAM occur but can
also happen when a swarm.IPAM and vendor/.../swarm.IPAM clollide.

Breaking changes:
1. (I)ContainerOperations.cs now uses ContainerExecStartParameters in place of ExecConfig that was a bug in the model originally.
   Fixed a typo in ContainerRestartParameters type.
2. (I)NetworkOperations.cs new returns a list of NetworkResponse type for `List` which is identical to the single type for inspect.


jterry75:
First step to #262 when I found out the models for swarm were wrong even in the current version due to name collisions with for example mount.Driver and swarm.Driver

galvesribeiro:
That explain some issues I had with mounting disks in the past. Great!

PR #277, opened 2018-01-12
RASUtil.GetCertFromPEMFiles fails with an IndexOutOfBoundsException

lanwin:
RASUtil.GetCertFromPEMFiles fails with an IndexOutOfBoundsException when pem file contains carriage returns.

Carriage Returns working perfectly fine in DockerTLS.

By the way. The readme.md suggest to use openssl to convert the pem files to PRX instead of using RASUtil.GetCertFromPEMFiles which is more practical.

### DIFF
--- a/src/Docker.DotNet.X509/RSAUtil.cs
+++ b/src/Docker.DotNet.X509/RSAUtil.cs
@@ -173,6 +173,10 @@ namespace Docker.DotNet.X509
                     index = 0;
                     while ((c = rdr.ReadChar()) != delim)
                     {
+                        if(c == '\r')
+                        {
+                            continue;
+                        }
                         line[index] = c;
                         index++;
                     }


galvesribeiro:
@jterry75 do you have anything against merging it? It is fine for me. Just concerned if there are not any limitation on older versions of .Net Framework where this would break it.

jterry75:
LGTM. @galvesribeiro - I think this would be safe in all versions

jterry75:
@lanwin - Thanks for the contribution!

lanwin:
Out of interest. Is there a reason why this impl dont uses a StreamReader instead of a BinaryReader?

jterry75:
Unless I am mistaken, the calls to `ReadUInt32(), ReadUInt16, ReadByte()` don't exist on the stream. We would have to do the casting from `ReadChar()` ourselves.

lanwin:
Yes of course but wouldn't the code be simple if the basic string parsing is done with am StreamReader and only the parsing of the Base64 is done with an BinaryReader?

jterry75:
I have never tried having two readers on the same stream before. If you want to give it a shot and see if it works I would be happy to simplify the code. They share the same underlying MemoryStream so its likely that it could work as long as we are careful around the close mechanisms.

PR #281, opened 2018-01-30
Fix to handle null web proxy settings.

kkmsft:
In cases where the Proxy is set to null, current code leads to an exception, which bubbles up to the caller. By adding a fix to check the null, we will return ProxyMode.None in such cases.


kkmsft:
Ping @jterry75

galvesribeiro:
Thanks @kkmsft 👍

PR #282, opened 2018-01-31
Updating to 3.125.1 release

jterry75:


jterry75:
Ok after this build will tag this release as the latest nuget. Can you verify the objects just to double check my work but I see the proper version now.

galvesribeiro:
@jterry75 I cloned your branch and Was looking for the Prune operations... Where are they?

Aren't they supposed to be on `ISystemOperations` ?

galvesribeiro:
I see the individual `PruneXXXAsync()` at each endpoint.

I could swear that we had a `ISystemOperations.PruneAsync()` which would represent the same as `docker system prune` and that would remove:

```
        - all stopped containers
        - all networks not used by at least one container
        - all dangling images
        - all build cache
```

If that is not the case,  we can address it in another PR.

jterry75:
```c#
IContainerOperations.PruneContainersAsync(...);
IImageOperations.PruneImagesAsync(...);
INetworkOpertions.PruneNetworksAsync(...);
```
What am I missing?

jterry75:
Hey we updated at the same time. There is no `SystemOperations.Prune` because there is no rest endpoint for that. We could certainly add one that would be a "client side call" to each of those endpoints but historically we have just exposed the rest endpoints.

jterry75:
@galvesribeiro - Are you wanting this before we update the release on NuGet I assume?

galvesribeiro:
@jterry75 no problem, second thought... I was going to use that on my Docker Orchestrator soon, but, you have a good point. The API don't provide it as a single endpoint, so we shouldn't add it, otherwise, we would later have to maintain it as extra work when the API changes.

Lets release it 💃 !

jterry75:
Haha deal.

jterry75:
In case you find yourself wanting it: https://github.com/jterry75/Docker.DotNet/tree/prune in some private code :).

galvesribeiro:
Hahahah thanks!

https://github.com/jterry75/Docker.DotNet/commit/c4bfe1e7a4ddfa238c363536dbe223440891091c#diff-c980c8dc3defefe7da98693e265e77c2R83

I would do that with `return Task.WhenAll(tasks)` ;)

jterry75:
Yea you win that one for sure.

Issue #271, opened 2017-11-29
No ways to retrieve stats stream with GetContainerStatsAsync() by using a IProgress<JSONMessage>


bcoste:
**Output of `dotnet --info`:**
Product Information:
 Version:            2.0.3
 Commit SHA-1 hash:  12f0c7efcc

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.15063
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.0.3\

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.3
  Build    : a9190d4a75f4a982ae4b4fa8d1a24526566c69df

**What version of Docker.DotNet?:**
3.125.0

**Steps to reproduce the issue:**
1. Obtain a DockerClient
2. Declare a progress to retrieve stats :
`IProgress<ContainerStatsResponse> progress = new Progress<ContainerStatsResponse>(stats => {
                            Console.WriteLine($"Stats: Read={stats.Read},PreRead={stats.PreRead},PidsStats={stats.PidsStats},BlkioStats={stats.BlkioStats},NumProcs={stats.NumProcs},StorageStats={stats.StorageStats},CPUStats={stats.CPUStats},PreCPUStats={stats.PreCPUStats},MemoryStats={stats.MemoryStats}");
                        });`

3. Try to call GetContainerStatsAsync() :
`await client.Containers.GetContainerStatsAsync(id, new ContainerStatsParameters(), progress, ct);`

**What actually happened?:**
Code does not compile because GetContainerStatsAsync() need a IProgress<JSONMessage>.
I don't know what to do with JSONMessage to get ContainerStatsResponse ?

**What did you expect to happen?:**
I would expect that GetContainerStatsAsync() requires a `IProgress<ContainerStatsResponse> progress` argument instead of `IProgress<JSONMessage>`.

**Additional information:**

jterry75:
I think you are totally correct! Want to try the private I just pushed and let me know if that solves your issue?

bcoste:
Thanks you for your feedback!
As it's non blocking, I'll stay with the obsolete version until next release.

jterry75:
Yea I am going to check all of the stream API's. I think I implemented them all at the same time and didn't realize they all didn't use the JSONMessage for progress. I will look into this and get an updated release out. Thanks!

sandersaares:
[...]

PR #287, opened 2018-02-15
Fixes an issue with GetContainerStatsAsync

jterry75:
Changes the response type from the stream to a ContainerStatsResponse and not JSONMessage like most of the progress API's use.

Resolves: #271


jterry75:
@galvesribeiro - Still awaiting confirmation on this so don't merge yet.

PR #289, opened 2018-02-27
Do not leak CancellationTokenSources

sandersaares:
DockerClient leaked objects if provided with both a timeout and a custom `CancellationToken`. It no longer does this.

The problem was that the created `CancellationTokenSource` was never disposed of, so it kept being referenced by the parent `CancellationToken` it was linked to. By disposing of the CTS after use the link is broken and the object can be collected.


jterry75:
Great fix! LGTM

sandersaares:
Do you have an estimate on when this fix might get into a NuGet release?

PR #292, opened 2018-03-05
Added support for network-sharing in ImageBuildParameters 

ptrkrnstnr:
(--network command)


galvesribeiro:
Thanks @ptrkrnstnr but I guess all the .generated.cs files come from the Go codegen so I think it would be better if @jterry75 generate it instead of manually change those files...

jterry75:
@galvesribeiro @ptrkrnstnr - Mind taking a look at #293?

jterry75:
We accepted #293 instead.

PR #293, opened 2018-03-06
Update ImageBuildParameters to latest set of options

jterry75:
Adds support for new options in ImageBuildParameters to reflect what docker
has added recently.

Signed-off-by: Justin Terry (VM) <juterry@microsoft.com>


jterry75:
@ptrkrnstnr - Thank you for the contribution in #292. Per @galvesribeiro I updated the model to generate this. Can you verify the fix because there was a minor discrepancy with your code. Docker makes this look like a `string` parameter not a `[]string` so are you sure it should have been an array?

galvesribeiro:
@ptrkrnstnr Once you confirm that will work work for you, we merge it.

ptrkrnstnr:
@galvesribeiro I included the new file in my project, but I just tested the NetworkMode property. This one is working fine for me. Can't tell you anything about the other new properties.

PR #296, opened 2018-03-23
Always report error response body

sandersaares:
This PR modifies the "make exception from error response" code to always read the response body in case of errors, as there is valuable information in there. When there is no error, the response body is not touched, as before.

Before this PR:

```
Docker API responded with status code=InternalServerError, response=
```

That error says nothing useful.

After this PR:

```
Docker API responded with status code=InternalServerError, response={"message":"Get https://registry.xx.com/v2/xx/xx-xx/manifests/cb: unauthorized: authentication required"}
```

Much better.

### DIFF
--- a/src/Docker.DotNet/DockerApiException.cs
+++ b/src/Docker.DotNet/DockerApiException.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿
+using System;
 using System.Net;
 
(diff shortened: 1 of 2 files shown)


galvesribeiro:
Thank you @sandersaares 

I never noticed that :)

jterry75:
Great fix thank you!

PR #297, opened 2018-03-24
Docker service test

abbgrade:
Hello,

this is a very simple test, that tests if the docker service is running, that explains why the other tests are failing with timeout exceptions.

Thanks,
Steffen

### DIFF
--- a/test/Docker.DotNet.Tests/ISystemOperations.Tests.cs
+++ b/test/Docker.DotNet.Tests/ISystemOperations.Tests.cs
@@ -16,6 +16,15 @@ namespace Docker.DotNet.Tests
         }
 
         [Fact]
+        public void DockerService_IsRunning()
+        {
+            using (var sc = new System.ServiceProcess.ServiceController("Docker"))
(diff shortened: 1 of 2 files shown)


test/Docker.DotNet.Tests/ISystemOperations.Tests.cs
```diff
@@ -15,6 +17,15 @@ public ISystemOperationsTests()
             _client = new DockerClientConfiguration(new Uri("npipe://./pipe/docker_engine")).CreateClient();
         }
 
+        [Fact]
+        public void DockerService_IsRunning()
+        {
+            var services = ServiceController.GetServices();
```
jterry75: Only one question. Why did you remove the:
```C#
using (var services = ...)
{
  //...
}
```
`Component` is disposable so shouldn't we do that?

abbgrade:
You are right. Its better to use `using` here.

jterry75:
Thank you for the contribution!

PR #298, opened 2018-03-25
Adding few operations' support for docker plugins api

prashantbhutani90:


src/Docker.DotNet/Models/PluginConfigInterface.Generated.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace Docker.DotNet.Models
+{
+    [DataContract]
+    public class PluginConfigInterface // (types.PluginConfigInterface)
+    {
+        [DataMember(Name = "Socket", EmitDefaultValue = false)]
+        public string Socket { get; set; }
+
+        [DataMember(Name = "Types", EmitDefaultValue = false)]
+        public IList<PluginInterfaceType> Types { get; set; }
```
prbhutan: This doesn't work for log drivers. The "Types" return for log driver is a string instead of PluginInterfaceType object.
Any ideas how to make sure this works for log drivers as well?

src/Docker.DotNet/Endpoints/IPluginOperations.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// </summary>
+        /// <remarks>
+        /// docker plugin pull
+        ///
+        /// HTTP POST /plugins/pull
+        ///
+        /// 204 - No error.
+        /// 500 - Server error.
+        /// </remarks>
+        Task InstallPluginAsync(IList<PluginInstallRequestBodyParameters> body, PluginInstallParameters parameters, IProgress<JSONMessage> progress, CancellationToken cancellationToken = default(CancellationToken));
```
jterry75: The other api's do the `body` part as an entry on the `parameters` object. Just for consistency we should move it.

src/Docker.DotNet/Endpoints/IPluginOperations.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// <remarks>
+        /// docker plugin ls
+        ///
+        /// HTTP GET /plugins
+        ///
+        /// 200 - No error.
+        /// 500 - Server error.
+        /// </remarks>
+        Task<IList<PluginListResponse>> ListPluginsAsync(PluginListParameters parameters, CancellationToken cancellationToken = default(CancellationToken));
+
```
jterry75: Mind adding an entry for `GetPluginPrivilegesAsync` and having it throw `NotImplementedException()`? Will make it super easy to see whats 'up-for-grabs'

jterry75: Also your generated `Plugin` but then you copied the type to `PluginListResponse`. Dont you want `IList<Plugin>` here?

src/Docker.DotNet/Endpoints/IPluginOperations.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// docker plugin disable
+        ///
+        /// HTTP POST /plugins/{name}/disable
+        ///
+        /// 200 - No error.
+        /// 404 - plugin is not installed.
+        /// 500 - Server error.
+        /// </remarks>
+        Task DisablePluginAsync(string name, CancellationToken cancellationToken = default(CancellationToken));
+    }
```
jterry75: And again, up to you but a `Upgrade/Create/Push/Configure` -> `NotImplementedException()` stub would be super helpful

src/Docker.DotNet/Endpoints/IPluginOperations.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// <remarks>
+        /// docker plugin inspect
+        ///
+        /// HTTP GET /plugins/{name}/json
+        ///
+        /// 200 - No error.
+        /// 404 - plugin is not installed.
+        /// 500 - Server error.
+        /// </remarks>
+        //Task<PluginInspectResponse> InspectPluginAsync(string id, CancellationToken cancellationToken = default(CancellationToken));
```
jterry75: Same here. Just have it throw `NotImplementedException()` if you don't want to finish it

src/Docker.DotNet/Models/PluginEnableParameters.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Runtime.Serialization;
+
+namespace Docker.DotNet.Models
+{
+    [DataContract]
+    public class PluginEnableParameters
+    {
+        [QueryStringParameter("timeout", false)]
+        public string Timeout { get; set; }
```
jterry75:
```C#
[QueryStringParameter("timeout", false)]
public uint TimeoutSeconds { get; set; }
```

src/Docker.DotNet/Models/PluginConfigInterface.Generated.cs
jterry75: Are you sure? Based on the spec I don't see how the go code would handle that case either. This is the correct generated code for the object type `PluginInterfaceType` which would be deserialized from json `object` not string.

prbhutan: Yes, I am sure,
I am implementing the API for supporting my custom log driver plugin. When I tried to list all the plugins, the code failed saying we cannot convert string to PluginInterfaceTypes object.
On using IList<string> the API worked fine.

jterry75: Can you paste what the JSON array is thats being returned in raw text? I would love to see what the daemon is sending.

prbhutan: [{"Config":{"Args":{"Description":"","Name":"","Settable":null,"Value":null},"Description":"Send log messages to elasticsearch","DockerVersion":"17.05.0-ce","Documentation":"-","Entrypoint":["/usr/bin/docker-log-elasticsearch"],"Env":[{"Description":"Set log level to output for plugin logs","Name":"LOG_LEVEL","Settable":["value"],"Value":"info"}],"Interface":{"Socket":"elasticsearchlog.sock","Types":["docker.logdriver/1.0"]},"IpcHost":false,"Linux":{"AllowAllDevices":false,"Capabilities":null,"Devices":null},"Mounts":null,"Network":{"Type":"host"},"PidHost":false,"PropagatedMount":"","User":{},"WorkDir":"","rootfs":{"diff_ids":["sha256:bd7b32cc16b423b23e33f6c770b5ef044f78af8b250f79d8e9c818bc07745394"],"type":"layers"}},"Enabled":true,"Id":"5216a226888623f939799f4ee640c88a399a621906a2c7f38e255170cabd738e","Name":"elasticsearch:latest","PluginReference":"docker.io/rchicoli/docker-log-elasticsearch:latest","Settings":{"Args":[],"Devices":[],"Env":["LOG_LEVEL=info"],"Mounts":[]}}]

jterry75: Here is the culprit: [plugin_response.go](https://github.com/moby/moby/blob/3a633a712c8bbb863fe7e57ec132dd87a9c4eff7/api/types/plugin_responses.go#L44). They have cheated :)! You have two options. Either change specgen.go to customize the "Types" field to be a string (easiest option) or you will need to build the JSON.Net Serialize/Deserialize code for this C# generated type that does the same .ToString()/Parse logic. Make sense?

prbhutan: How do I make "Types" field a string in specgen.go?

prbhutan: Thanks. Will make the change soon.

src/Docker.DotNet/Endpoints/PluginOperations.cs
```diff
@@ -0,0 +1,115 @@
...
+            return StreamUtil.MonitorStreamForMessagesAsync(
+                this._client.MakeRequestForStreamAsync(this._client.NoErrorHandlers, HttpMethod.Post, $"plugins/pull", queryParameters, data, null, CancellationToken.None),
+                this._client,
+                cancellationToken,
+                progress);
+        }
+
+        public async Task<IList<Plugin>> ListPluginsAsync(PluginListParameters parameters, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            if (parameters == null)
```
jterry75: This one should be optional as there are no required parameters

src/Docker.DotNet/Endpoints/PluginOperations.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        public Task RemovePluginAsync(string name, PluginRemoveParameters parameters, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            if (string.IsNullOrEmpty(name))
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            if (parameters == null)
```
jterry75: Same should be optional.

src/Docker.DotNet/Endpoints/PluginOperations.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        public Task EnablePluginAsync(string name, PluginEnableParameters parameters, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            if (string.IsNullOrEmpty(name))
+            {
+                throw new ArgumentNullException(nameof(name));
+            }
+
+            if (parameters == null)
```
jterry75: Same should be optional.

tools/specgen/specgen.go
```diff
@@ -140,6 +140,12 @@ var typesToDisambiguate = map[string]*CSModelType{
 	typeToKey(reflect.TypeOf(types.NetworkDisconnect{})):     {Name: "NetworkDisconnectParameters"},
 	typeToKey(reflect.TypeOf(types.NetworksPruneReport{})):   {Name: "NetworksPruneResponse"},
 	typeToKey(reflect.TypeOf(types.NetworkResource{})):       {Name: "NetworkResponse"},
+	typeToKey(reflect.TypeOf(types.PluginConfigInterface{})): {
+		Name: "PluginConfigInterface",
+		Properties: []CSProperty{
+			CSPropertyName{Name: "Types", Type: CSType{"System.Collections.Generic", "IList<string>", false}},
+		},
+	},
```
jterry75: This looks good did you forget to re-run update? I am surprised to still see the PluginInterfaceType.Generated.cs

prashantbhutani90: I tried to install go compiler on my windows machine. Apparently, the site was unreachable yesterday, and I thought this could be an independent change.

Issue #306, opened 2018-04-19
Timeout field of PluginEnableParameter should be a string not long?


prashantbhutani90:
I installed a plugin and tried to enable it. With Timeout = 0, docker throws strconv.Atoi exception saying invalid syntax.
If I directly use the source code with Timeout field set to string, and passing Timeout = "0" succeeds.

jterry75:
Hmm, from what I can tell the documentation from [Docker](https://docs.docker.com/engine/api/v1.26/#operation/PluginEnable), the [typedef](https://github.com/moby/moby/blob/9769ef333f2af24b30fed0dd7b00384b2df3b953/api/types/configs.go#L50) and the handler all say `int`. 

Could you do me a favor and post the `json` from running your daemon with the `-D` flag while using the `docker.exe` to enable the plugin and when using `Docker.DotNet` to enable the plugin? I fail to see how this should be treated as a `string` and just want to make sure we are fixing the right thing. The `json` should show us the differences in how we are serializing vs how the `docker cli` is.

Thanks!

prashantbhutani90:
Apparently, docker complains only if Timeout is set to 0 or not set at all.
For non-zero integer value for Timeout, plugin enable API works fine.

jterry75:
I think this is a bug in docker. The API documentation has marked that this is an optional, integer with default value of 0. However the router and client `ALWAYS` specify a `string` therefore making this required. 

I think the right fix is to make [modeldefs.go](https://github.com/Microsoft/Docker.DotNet/blob/2aa85f3e3d14b6d44fe2afdc07dc35fa70b2d357/tools/specgen/modeldefs.go#L289)
Change to:
```
// PluginEnableParameters for POST /plugins/(name)/enable
type PluginEnableParameters struct {
	Timeout string `rest:"query,Timeout,required,\"0\""`
}
```

Regenerate the code and that should change PluginEnableParameters.cs

jterry75:
I see your PR is still up. Did you have a chance to try this alternate way out?

PR #307, opened 2018-04-19
Timeout property of PluginEnableParameters should be string. 

prashantbhutani90:
Fixes issue #306


prashantbhutani90:
To be clear, it fixes the issue after we regenerate the cs files from go.

PR #312, opened 2018-05-21
Allow default values of Value Types to be send to docker API as query string parameters

andrey-skat:
For now we can't send explicitly `false` or `0` to docker API, because default Type values are removed from query parameters. Those values works without bugs by the fact that almost all Boolean parameters in Docker API are `false` by default. 

But I think that creating code on the assumption that `false` always be default value is not good idea. 
It can lead to painful debug in future. So I think we shouldn't remove those parameters from query string.

### DIFF
--- a/src/Docker.DotNet/QueryString.cs
+++ b/src/Docker.DotNet/QueryString.cs
@@ -42,7 +42,7 @@ namespace Docker.DotNet
                 }
 
                 // Serialize
-                if (attribute.IsRequired || !IsDefaultOfType(value))
+                if (value != null)
                 {
                     var keyStr = attribute.Name;
                     string[] valueStr;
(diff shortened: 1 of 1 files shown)


jterry75:
This is a very common practice in the world of REST. Why send data that is the default value of its type? We have a facility to enable this by making the value optional. Consider the following:

```go
// Go Service Side
type T struct {
    Test bool `json:",omitempty"`
}
```
```json
{
    "Test": false
}
```
By contract this is allowing for the case where if `T.Test` is `false` what is the point of including it in json?

However, if you *really* want to send it from the `Docker.DotNet` client we can do it this way:
```c#
// C#
class T
{
    bool? Test;
}
```
Now its `default value` is `null` so the marshaller will skip writing it if the client has set it to any boolean. However if you set it to `t.Test = false` it is no longer its default value and will be forcibly marshalled even though the service assumed `false` by default. Does this make sense?

andrey-skat:
You are right. But even though I'm not too familiar with the library code, I think that now this part is not obvious. Because of default Type Value there equals to default value in Docker API. But in `ContainerStatsParameters`  Stream's default is `true` and parameter `false` is not removed just because it required.

It's not critical, but I think the reason why parameter goes to query string should be more clear. Now the reasons: it's required or it's not default type value for c# (which implicitly not default value for docker api). Instead it should be: not default value for docker api (or not passed at all, i.e. `null`)

I can close PR, if it is not needed :)

jterry75:
Ahh. I see the issue now. Thanks for providing `ContainerStatsParameters` that example is perfect to identify the issue. Let me think about this for a bit and I will decide if this is the best fix of if we have other options. Leave the PR for now.

PR #316, opened 2018-05-30
Give ImageDeleteParameters-PruneChildren a better name

JasonStein:
Since PruneChildren = true actually means no prune children in the code when sending request to Docker daemon and this parameter is called "--no-prune" in Docker API, so Let us give it a better name here.


jterry75:
@JasonStein - Looks great! But you forgot to run the updates on the generated code that updates the .cs file associated with that.

JasonStein:
Thanks for the remind, just figure out I need to manually build and run the specGen :) Updated generated model code in the latest commit.

jterry75:
LGTM. Thanks for the contribution!

PR #323, opened 2018-06-27
Enable support for UNIX sockets on all runtimes

### LINKED ISSUE #322: 'Unknown URL scheme unix' Exception when running with mono on Linux
**Output of `dotnet --info`:**

```
 $ dotnet --info
.NET Command Line Tools (2.1.201)

Product Information:
 Version:            2.1.201
 Commit SHA-1 hash:  7932dc6179

Runtime Environment:
 OS Name:     ubuntu
 OS Version:  16.04
 OS Platform: Linux
 RID:         ubuntu.16.04-x64
 Base Path:   /home/phil/.dotnet/sdk/2.1.201/

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.7
  Build    : 2d61d0b043915bc948ebf98836fefe9ba942be11
```

**What version of Docker.DotNet?:**

```
3.125.2
```

**Steps to reproduce the issue:**
1.  Attempt to use the library to connect to the docker daemon on linux, building with mono (in a project targeting `net471`): `var dockerClient = new DockerClientConfiguration("unix:///var/run/docker.sock").CreateClient();
2. Exception will be thrown.

**What actually happened?:**
An Exception with the message "Unknown URL scheme unix" was thrown.

**What did you expect to happen?:**
The connection to the docker daemon was made successfully, just as if building and running a dotnet core `netcoreapp2.0` project.

alastairs:
.NET Framework applications are capable of connecting to UNIX sockets, and System.Net.Sockets ships as part of the .NET Framework. The Mono runtime supports .NET Framework applications on a cross-platform basis. I can't see a reason right now to not support UNIX socket URLs on all runtimes. 

Closes #322.

### DIFF
--- a/src/Docker.DotNet.BasicAuth/Docker.DotNet.BasicAuth.csproj
+++ b/src/Docker.DotNet.BasicAuth/Docker.DotNet.BasicAuth.csproj
@@ -3,7 +3,7 @@
     <IsPackable>true</IsPackable>
     <PackageId>Docker.DotNet.BasicAuth</PackageId>
     <Description>Docker.DotNet.BasicAuth is a library that allows you to use basic authentication with a remote Docker engine programmatically in your .NET applications.</Description>
-    <TargetFrameworks>netstandard2.0;netstandard1.6;net46;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard1.6</TargetFrameworks>
     <AssemblyName>Docker.DotNet.BasicAuth</AssemblyName>
     <PackageIconUrl>https://camo.githubusercontent.com/fa6d5c12609ed8a3ba1163b96f9e9979b8f59b0d/687474703a2f2f7765732e696f2f566663732f636f6e74656e74</PackageIconUrl>
     <PackageLicenseUrl>https://github.com/Microsoft/Docker.DotNet/blob/master/LICENSE</PackageLicenseUrl>
--- a/src/Docker.DotNet.X509/Docker.DotNet.X509.csproj
+++ b/src/Docker.DotNet.X509/Docker.DotNet.X509.csproj
@@ -3,7 +3,7 @@
     <IsPackable>true</IsPackable>
     <PackageId>Docker.DotNet.X509</PackageId>
     <Description>Docker.DotNet.X509 is a library that allows you to use certificate authentication with a remote Docker engine programmatically in your .NET applications.</Description>
-    <TargetFrameworks>netstandard2.0;netstandard1.6;net46;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard1.6</TargetFrameworks>
     <AssemblyName>Docker.DotNet.X509</AssemblyName>
     <PackageIconUrl>https://camo.githubusercontent.com/fa6d5c12609ed8a3ba1163b96f9e9979b8f59b0d/687474703a2f2f7765732e696f2f566663732f636f6e74656e74</PackageIconUrl>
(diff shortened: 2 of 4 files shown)


alastairs:
a7d6d4b reduces the targeted frameworks to .NET Standard 1.6 and .NET Standard 2.0. This maintains support in the library for .NET 4.6.1 and higher, but drops support for .NET 4.5 which does not implement a version of .NET Standard containing UnixDomainSocketEndPoint. I realise this might be a showstopper for you :)

jterry75:
We cant remove support for .NET45 and 46

Can't we just make the define:
```C#
#if !(NET45 || NET46)
```
This allows us to keep the target frameworks as is and also support all of the latest frameworks right?

alastairs:
Well, .NET 4.6 would still be supported via .NET Standard 1.6 and 2.0; only .NET 4.5 support would be dropped, but I guess that’s not an option either. 

The issue with inverting the `#if` is that the logic remains the same and so the bug remains 😊 The crux of the matter is that projects targeting .NET Framework 4.7 [seem to] pick up the net45/net46 binary, presumably the latter, rather than either of the .NET Standard binaries. There is no way to override this behaviour with NuGet, although I’m looking at a workaround with paket. 

There is an implementation of the necessary  UNIX socket class in the Mono Project, which could be copied in as a source file to maintain compatibility with .NET 4.5. It may also be possible to pull it in from .NET Core, I guess?

jterry75:
@jstarks - Can we drop support for `net45` or `net46`?

alastairs:
Any update @jterry75 @jstarks?

jterry75:
Hey @alastairs - Thought more about this. Does this also solve your problem? [jterry75/Docker.DotNet:c55ecec](https://github.com/jterry75/Docker.DotNet/commit/c55cec25cc4bde58d3c34c7350c98d35d9577e9a). It seems like adding the `.NET471` TFM and putting that on Nuget would allow people targeting `NET471` to also use Unix sockets by default because the binary wouldn't fall back to `NET46`.

Want to give this a shot and let me know your feedback?

(I could do `.NET47` as well but they changed the `Socket.ConnectAsync` pattern so that was a harder proof of concept to show you. I understand that this would require users to to target `.NET471` and newer).

alastairs:
@jterry75 Yup, we can definitely try that out. Where can we pick up the NuGet package from?

jterry75:
@alastairs - Sorry didn't see you asked me for something. [artifcats](https://ci.appveyor.com/project/WindowsVirtualization/docker-dotnet/build/3.125.2.111-cfjjppxi/artifacts). If you can give this a shot I will put out a new build of Docker.DotNet today with this change as well as all the others since last release. Please let me know, I wan't to make sure this change its in as well.

0xced:
I tried to run on mono/net471 with your patch ([terry75/Docker.DotNet:c55ecec](https://github.com/jterry75/Docker.DotNet/commit/c55cec25cc4bde58d3c34c7350c98d35d9577e9a)) and it works fine.

Note that if both this pull request and #340 are merged, there will be a compilation conflict because of `<PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />` introduced in #340. That should be easy to fix by only adding that package reference when targeting net45 and net46.

Neuroforge:
Is this going to get merged? Needs this to connect to docker on MacOS.

jterry75:
@Neuroforge - As stated above we cant remove `net45` or `net46`? Are you targeting `net47*`?

alastairs:
Closing this PR as I'm no longer working on the code base that exhibited the problem.

PR #340, opened 2018-08-31
Automatic detection of Windows / Unix

0xced:
* Added a new DockerClientConfiguration constructor that automatically detect the current platform and chooses the local docker endpoint Uri using `npipe://./pipe/docker_engine` on Windows and `unix:/var/run/docker.sock` otherwise.

* Updated documentation to reflect that new feature

* Updated the tests to use this new constructor. The tests can now run on Linux, macOS and Windows.

### DIFF
--- a/src/Docker.DotNet/Docker.DotNet.csproj
+++ b/src/Docker.DotNet/Docker.DotNet.csproj
@@ -19,6 +19,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
     <PackageReference Include="System.ValueTuple" Version="4.4.0" />
   </ItemGroup>
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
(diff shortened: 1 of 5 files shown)


jterry75:
LGTM. But can we keep the documentation to show how to author the `npipe:// and unix://` address configurations. Just make it say something like: For a custom endpoint `unix://` `npipe://` `tcp://` etc. People ask about that frequently.

src/Docker.DotNet/Docker.DotNet.csproj
```diff
@@ -19,6 +19,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
+    <PackageReference Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
```
galvesribeiro: Shouldn't this package be included under a `Condition` to check the target framework for `net46` and `net45`?

jterry75: Looks like 4.3.0 package has built in support for net45/46/netstandard1.1 so it works to include it on all here.

galvesribeiro:
Thanks @0xced!

0xced:
Thanks for merging this. Unfortunately, the latest published version is 3.125.2 which does not support  a parameterless constructor for `DockerClientConfiguration`. So the documentation is now out of sync with the actual published package. 😕

PR #343, opened 2018-09-18
A cancellation doesn't work in some cases and a deadlock may occure. …

suneption:
…I added an additional cancellation.


YakhontovYaroslav:
@jterry75 
I`ve encountered this issue too when tried to read event from Docker for Windows (18.03 if that matters). 
My understanding is it is hanging somewhere in native code inside ReadLineAsync, trying to find end of line delimiter, but if there is no docker events at all it cannot return to managed world and react to token cancellation.

Issue #261, opened 2017-09-25
Move exec operations from ContainerOperations


jterry75:
To better match the API the 4+ methods around Exec should be moved from IContainerOperations to IExecOperations. This will be a breaking change but a very easy one to upgrade as its a move operation.

PR #346, opened 2018-10-04
Changes for Move exec operations from ContainerOperations

GiancarloLelli:
Hi teams
I saw #261 that was up for grabs and I jumped in.

Cheers
Giancarlo Lelli


jterry75:
Awesome thanks! LGTM

PR #350, opened 2018-10-15
Add filter capability to the ListService swarm method

L3Z4:
Following the documentation available here: https://docs.docker.com/engine/api/v1.32/#tag/Service, I simply added the possibility to the "`ListServicesAsync`" method of the Swarm part of the client to be called with a new parameter (`FilterServiceParameters`).

I also add unit tests about my new implementation of "`IQueryString`": "`FilterServiceQueryString`".

I tested my work using my own swarm cluster, with the class `ISystemOperationTests`. I pushed this code, but it depends on the cluster on which you run.


galvesribeiro:
Thanks for the contribution @L3Z4 

@jterry75 any concerns?

jterry75:
In general I would prefer we follow the same pattern we have for the rest of the code. We would add a `ServicesListParameters` which would have a `Filters` property. I think that we could accomplish this more generically by having a `ServiceFilters : IDictionary<string, bool>` which just adds the type safe `getters/setters` for the specific filters it allows. And the the `Filters` property would be of this `ServicesFilters` type. We could expand that idea to all of the API's that have filters ya know?

jterry75:
But I am not specifically opposed to this change in any way just wondering if we should make the solution for other areas as well. Feel free to say no

galvesribeiro:
Humm... Looking that way, yeah, make sense... Other APIs also support filtering and makes sense to have it in a generic-ish way...

@L3Z4 Do you mind changing it as suggested by @jterry75 ?

Thanks!

L3Z4:
Hi, 

Thanks for th review. I agree to make this changes, it allows the design to be open to new filter not designed yet. Furthermore, it allow me to check the mode limitation inside the setter and not in the constructor. I'm glad 👍 
I'll update the PR this moring.

galvesribeiro:
@jterry75 ?

src/Docker.DotNet/Models/ServiceFilter.cs
```diff
@@ -0,0 +1,62 @@
...
+            set
+            {
+                if (ContainsKey("label"))
+                    this["label"] = value;
+                else
+                    Add("label", value);
+            }
+        }
+        /// <summary> "replicated"|"global" </summary>
+        public string Mode
```
jterry75: Why not make this an enum with these values and do the parsing in the get/set but store in the dictionary as string so that we can have the type safety

src/Docker.DotNet/Models/ServiceFilter.cs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    public class ServiceFilter : Dictionary<string, string>
+    {
+        public string Id
+        {
+            get => this["id"];
+            set
+            {
+                if (ContainsKey("id"))
```
jterry75: I'm surprised you have to make this check. Why cant set just be `this["id"] = value;`

src/Docker.DotNet/Models/ServiceFilter.cs
```diff
@@ -0,0 +1,62 @@
+using System;
+using System.Collections.Generic;
+
+namespace Docker.DotNet.Models
+{
+    public class ServicesListParameters
+    {
+        public ServiceFilter Filters { get; set; }
```
jterry75: I was hoping that these would use `[QueryStringParameter("filters", false, typeof(MapQueryStringConverter))]` and you would enlighten `MapQueryStringConverter` to process these so we could avoid adding `FilterServiceQueryString`.

L3Z4:
@jterry75 Thanks for the review. At the begining, I did not understand the meaning of the `QueryStringParameter` attribute. I think it needs some documentation.
I added unit test to ensure the functionnality of my code, but i did not it with a real swarm cluster.

L3Z4:
Sorry about the force push, I would like to ensure the continuity of the commits. I did not know it would delete your changes requests.

src/Docker.DotNet/Endpoints/SwarmOperations.cs
```diff
@@ -130,6 +129,15 @@ async Task<IEnumerable<SwarmService>> ISwarmOperations.ListServicesAsync(Cancell
             return this._client.JsonSerializer.DeserializeObject<SwarmService[]>(response.Body);
         }
 
+        async Task<IEnumerable<SwarmService>> ISwarmOperations.ListServicesAsync(ServicesListParameters parameters, CancellationToken cancellationToken)
+        {
+            IQueryString queryParameters = new QueryString<ServicesListParameters>(parameters);
```
jterry75: Can you validate this against null and skip the query parameters if it is. Then update the original `ListServicesAsync` to just call this so we don't have two impls.

src/Docker.DotNet/Models/ServiceFilter.cs
jterry75: Change the order of your getter/setter and use an enum.
```C#
public ServiceCreationMode Mode
{
    get
    {
        ServiceCreationMode mode;
        Enum.TryParse(this["mode"], out mode);
        return mode;
    }
    set
    {
        this["mode"] = value.ToString()
    }
}
```

src/Docker.DotNet/Models/ServiceFilter.cs
```diff
@@ -0,0 +1,52 @@
...
+                this["mode"] = value;
+            }
+        }
+        public string Name
+        {
+            get => this["name"];
+            set => this["name"] = value;
+        }
+
+        private enum ServiceCreationMode
```
jterry75: Make it public so that consumers can use it in a type safe way.

src/Docker.DotNet/QueryString.cs
```diff
@@ -1,154 +1,182 @@
...
+        /// <returns></returns>
+        public string GetQueryString()
+        {
+            return string.Join("&",
+                        _data.Select(
+                            v => $"{Uri.EscapeUriString(_key)}={Uri.EscapeDataString(v)}"));
+        }
+    }
+
+    //internal class FilterServiceQueryString : IQueryString
```
jterry75: Can we remove?

src/Docker.DotNet/Endpoints/SwarmOperations.cs
L3Z4: Agreed.

src/Docker.DotNet/Models/ServiceFilter.cs
L3Z4: Agreed.

src/Docker.DotNet/QueryString.cs
L3Z4: Of course, my bad !

test/Docker.DotNet.Tests/ISystemOperations.Tests.cs
```diff
@@ -90,6 +90,19 @@ public async Task MonitorEventsAsync_Succeeds()
             await _client.Images.DeleteImageAsync($"{repository}:{tag}", new ImageDeleteParameters());
         }
 
+        [Fact]
+        public async Task GetServicesAsync_Succeeds()
```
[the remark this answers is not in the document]
L3Z4: Done !

test/Docker.DotNet.Tests/ServiceFilterTests.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Docker.DotNet.Tests
+{
+    public class ServiceFilterTests
+    {
+        [Fact]
+        public void ServiceFilter_AddInvalidMode()
+        {
+            var sf = new ServiceFilter();
+            Assert.Throws<Exception>(() => sf.Mode = "invalid");
```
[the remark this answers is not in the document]
L3Z4: I'm not sure. It would imply an unexpected behavior from a mistake made by the developer using this lib.

L3Z4:
Is these changes acceptables ?

L3Z4:
I've missed your point about the enum. 
I've just made the changes.

jterry75:
Looks great! @galvesribeiro - PTAL

galvesribeiro:
Thanks @L3Z4!

PR #351, opened 2018-10-15
A cancellation doesn't work in some cases and app hangs.

gurustron:
I've played a little bit with an issue raised [here](https://github.com/Microsoft/Docker.DotNet/pull/343) it seems that without setting NamedPipeClientStream to PipeOptions.Asynchronous after disposing stream it endlessly waits in PipeStream.Read for local Windows connection.

Steps to repro:

create console app (NET71, .NETCORE) on Windows machine with empty Docker. 

        public static async Task Main(string[] args)
        {
            var cts = new CancellationTokenSource();

            var _client = new DockerClientConfiguration().CreateClient();
            cts.CancelAfter(3000);
            cts.Token.Register(() => { Console.WriteLine("Cancel happened"); });
            var task = _client.System.MonitorEventsAsync(new ContainerEventsParameters(), new Progress(), cts.Token);
            await task.ConfigureAwait(false);
            Console.WriteLine("finished");
            Console.ReadKey();
        }
        
        class Progress : IProgress<JSONMessage>
        {
            internal Action<JSONMessage> _onCalled;

            void IProgress<JSONMessage>.Report(JSONMessage value)
            {
                _onCalled(value);
            }
        }

### DIFF
--- a/src/Docker.DotNet/DockerClient.cs
+++ b/src/Docker.DotNet/DockerClient.cs
@@ -75,7 +75,11 @@ namespace Docker.DotNet
                         // NamedPipeClientStream handles file not found by polling until the server arrives. Use a short
                         // timeout so that the user doesn't get stuck waiting for a dockerd instance that is not running.
                         var timeout = 100; // 100ms
-                        var stream = new NamedPipeClientStream(serverName, pipeName);
+                        var stream = new NamedPipeClientStream(
+                            serverName,
+                            pipeName,
+                            PipeDirection.InOut,
+                            PipeOptions.Asynchronous);
                         var dockerStream = new DockerPipeStream(stream);
 
 #if NET45
--- a/src/Docker.DotNet/Microsoft.Net.Http.Client/BufferedReadStream.cs
+++ b/src/Docker.DotNet/Microsoft.Net.Http.Client/BufferedReadStream.cs
@@ -160,9 +160,11 @@ namespace Microsoft.Net.Http.Client
             {
                 _bufferOffset = 0;
                 _bufferCount = await _inner.ReadAsync(_buffer, _bufferOffset, _buffer.Length, cancel).ConfigureAwait(false);
+
+                ThrowIfDisposed();
+
                 if (_bufferCount == 0)
                 {
-                    ThrowIfDisposed();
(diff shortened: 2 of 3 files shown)


marcelltoth:
I'm running into the same issue.

It looks like microsoft kind of ignores this repo, in the meanwhile does anyone have a good workaround? (Other than just keeping all my tasks alive, which does not sound great.)

dgvives:
Hello @gurustron 
I'm going through all unmerged PR, my intention is to create a PR with code compatible with latest Docker API. 
There are outdated models in current definition and this is causing automatic tests to fail within the CI/CD pipeline, causing PR to not be even reviewed.

Would you check if MonitorEventsAsync is working? I haven't managed to

jterry75:
@dgvives - Unrelated to this PR but if you are going to do that I would prefer that we just remove tools/specgen anymore and just manually update the models. Docker changed how all of this worked long ago when the broke the branches into EE/CE and made specgen less than helpful. I'm all for updating the models I just think that specgen is not worth the effort anymore. I no longer have the time to really contribute here but I am happy to merge PR's if you want to do this work. Thanks!

galvesribeiro:
Yeah, that is what I’m working into. A client that doesn’t depends on Go anymore.

Sorry if it is taking too long. Life is being hard recently...

My idea is to generate things based on the yaml from the API with Roslyn.

Still working on it...

Get Outlook for iOS<https://aka.ms/o0ukef>

dgvives:
@jterry75 @galvesribeiro  Some of the models overridden on modeldefs.go are missing the required rest tag for correct model generation. I've updated some I found broken, but not all of them.

Code generation using specgen is what I've used. All the generated models are based on specgen details, as I think they should be at least for now. 

Continue on #480

gurustron:
@dgvives  I will take a look into it this week.

dgvives:
@gurustron I confirm these changes have been incorporated in the last merge, `MonitorEventsAsync` can be cancelled
Current solution is based on #343 using an extension method and throwing an exception whenever token gets cancellated. 

However, this behaviour is not the most appropriate as it produces slightly different results when running tests on different environments. I expect to submit a PR with a different approach to this, and include references to yours and others' work  as contributor

PR #357, opened 2018-11-05
Set NamedPipeClientSteam PipeOptions to asynchronous and make timeout configurable

sindhura:
The timeout for calling Connect on the NamedPipeClientStream used to communicate with the dockerd instance was hardcoded to 100 milliseconds. Made it so it can be configured in DockerClientConfiguration.
Also set the PipeOptions in the NamedPipeClientStream constructor to Asynchronous

### DIFF
--- a/src/Docker.DotNet/DockerClientConfiguration.cs
+++ b/src/Docker.DotNet/DockerClientConfiguration.cs
@@ -12,6 +12,10 @@ namespace Docker.DotNet
 
         public TimeSpan DefaultTimeout { get; internal set; } = TimeSpan.FromSeconds(100);
 
+        // NamedPipeClientStream handles file not found by polling until the server arrives. Use a short
+        // timeout so that the user doesn't get stuck waiting for a dockerd instance that is not running.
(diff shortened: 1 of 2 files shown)


src/Docker.DotNet/DockerClientConfiguration.cs
```diff
@@ -12,6 +12,10 @@ public class DockerClientConfiguration : IDisposable
 
         public TimeSpan DefaultTimeout { get; internal set; } = TimeSpan.FromSeconds(100);
 
+        // NamedPipeClientStream handles file not found by polling until the server arrives. Use a short
```
jterry75: Can we just have this read:
```C#
// The number of milliseconds to wait for the server to respond before the connection times out.
```
This is what the C# documentation says and its less confusing than the comment that was there before.

sindhura: Personally I did not want the comment there… I copied it from the previous location where the 100 ms was declared.
I’m inclined to remove the comment altogether if that is alright. The variable name is descriptive enough i think.

jterry75: Works for me.

PR #358, opened 2018-11-18
Services filter fixed. And service operations now throws dockerapiexc…

dud380:
Swarm operations throw DockerApiException on error. 
Service filter bug fixed
Tests updated
Tests added for service operations

### DIFF
--- a/src/Docker.DotNet/Endpoints/SwarmOperations.cs
+++ b/src/Docker.DotNet/Endpoints/SwarmOperations.cs
@@ -13,10 +13,9 @@ namespace Docker.DotNet
     {
         internal static readonly ApiResponseErrorHandlingDelegate SwarmResponseHandler = (statusCode, responseBody) =>
         {
-            if (statusCode == HttpStatusCode.ServiceUnavailable)
+            if (statusCode < HttpStatusCode.OK || statusCode >= HttpStatusCode.BadRequest)
             {
-                // TODO: Make typed error.
-                throw new Exception("Node is not part of a swarm.");
+                throw new DockerApiException(statusCode, responseBody);
             }
         };
 
@@ -46,21 +45,11 @@ namespace Docker.DotNet
         {
             var data = new JsonRequestContent<SwarmInitParameters>(parameters ?? throw new ArgumentNullException(nameof(parameters)), this._client.JsonSerializer);
             var response = await this._client.MakeRequestAsync(
-                new ApiResponseErrorHandlingDelegate[]
-                    {
-                        (statusCode, responseBody) =>
-                        {
-                            if (statusCode == HttpStatusCode.NotAcceptable)
-                            {
-                                // TODO: Make typed error.
-                                throw new Exception("Node is already part of a swarm.");
-                            }
-                        }
-                    },
+                new[] { SwarmResponseHandler },
                 HttpMethod.Post,
                 "swarm/init",
-                null,
-                data,
+                null, 
+                data, 
                 cancellationToken).ConfigureAwait(false);
 
             return response.Body;
@@ -70,13 +59,21 @@ namespace Docker.DotNet
         {
             if (string.IsNullOrEmpty(id)) throw new ArgumentNullException(nameof(id));
 
-            var response = await this._client.MakeRequestAsync(new[] { SwarmResponseHandler }, HttpMethod.Get, $"services/{id}", cancellationToken).ConfigureAwait(false);
+            var response = await this._client.MakeRequestAsync(
+                new[] { SwarmResponseHandler }, 
+                HttpMethod.Get, 
+                $"services/{id}", 
+                cancellationToken).ConfigureAwait(false);
             return this._client.JsonSerializer.DeserializeObject<SwarmService>(response.Body);
         }
 
         async Task<SwarmInspectResponse> ISwarmOperations.InspectSwarmAsync(CancellationToken cancellationToken)
         {
-            var response = await this._client.MakeRequestAsync(new[] { SwarmResponseHandler }, HttpMethod.Get, "swarm", cancellationToken).ConfigureAwait(false);
+            var response = await this._client.MakeRequestAsync(
+                new[] { SwarmResponseHandler }, 
+                HttpMethod.Get, 
+                "swarm", 
+                cancellationToken).ConfigureAwait(false);
             return this._client.JsonSerializer.DeserializeObject<SwarmInspectResponse>(response.Body);
         }
 
@@ -84,21 +81,11 @@ namespace Docker.DotNet
         {
             var data = new JsonRequestContent<SwarmJoinParameters>(parameters ?? throw new ArgumentNullException(nameof(parameters)), this._client.JsonSerializer);
             await this._client.MakeRequestAsync(
-                new ApiResponseErrorHandlingDelegate[]
-                {
-                    (statusCode, responseBody) =>
-                    {
-                        if (statusCode == HttpStatusCode.ServiceUnavailable)
-                        {
-                            // TODO: Make typed error.
-                            throw new Exception("Node is already part of a swarm.");
-                        }
-                    }
-                },
-                HttpMethod.Post,
-                "swarm/join",
-                null,
-                data,
+                new[] { SwarmResponseHandler }, 
+                HttpMethod.Post, 
+                "swarm/join", 
+                null, 
+                data, 
                 cancellationToken).ConfigureAwait(false);
         }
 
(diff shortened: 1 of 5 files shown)


src/Docker.DotNet/Endpoints/SwarmOperations.cs
```diff
@@ -13,10 +13,9 @@ internal class SwarmOperations : ISwarmOperations
     {
         internal static readonly ApiResponseErrorHandlingDelegate SwarmResponseHandler = (statusCode, responseBody) =>
         {
-            if (statusCode == HttpStatusCode.ServiceUnavailable)
+            if (statusCode < HttpStatusCode.OK || statusCode >= HttpStatusCode.BadRequest)
```
jterry75: This is what the default handler does. Should we just remove the custom handler?

Alternatively if you want to have a more 'typed' style exception you could have this read:

```C#
if (statusCode == HttpStatusCode.ServiceUnavailable)
{
    throw new DockerApiException(statusCode, "Node is not part of a swarm: " + responseBody);
}
else if (statusCode == HttpStatusCode.NotAcceptable)
{
    throw new DockerApiException(statusCode, "Node is already part of a swarm: " + responseBody);
}
```

jterry75: Or maybe just split it into a Positive/Negative match and update the functions as the API makes sense. Not sure if `ServiceUnavailable` or `NotAccepted` are ever returned from the functions for different meaning. Would need to look at the docs.

src/Docker.DotNet/Models/ServiceFilter.cs
```diff
@@ -10,24 +10,39 @@ public class ServicesListParameters
...
             set => this["label"] = value;
         }
-        public ServiceCreationMode Mode
+        public ServiceCreationMode[] Mode
         {
-            get => !Enum.TryParse(this["mode"], out ServiceCreationMode mode) ? ServiceCreationMode.Replicated : mode;
-            set => this["mode"] = value.ToString();
+            get
+            {
+                if (this["mode"].Length == 0)
```
jterry75: If this is an `array` filter I would prefer the default value is `null`.

test/Docker.DotNet.Tests/ISwarmOperationsTests.cs
```diff
@@ -2,30 +2,75 @@
...
+        private readonly string _testServiceId;
+        private static string _testServiceName = "docker-dotnet-test-service";
 
         public ISwarmOperationsTests()
         {
             _client = new DockerClientConfiguration().CreateClient();
+
+            _testServiceId = _client.Swarm.CreateServiceAsync(new ServiceCreateParameters()
+            {
+
```
jterry75: NIT; extra space

test/Docker.DotNet.Tests/ISwarmOperationsTests.cs
```diff
@@ -2,30 +2,75 @@
...
-            var services = await _client.Swarm.ListServicesAsync(new ServicesListParameters { Filters = new ServiceFilter { Id = "pr6264hhb836" } }, CancellationToken.None);
+            var services = await _client.Swarm.ListServicesAsync(new ServicesListParameters { Filters = new ServiceFilter { Name = new string[] { _testServiceName } } }, CancellationToken.None);
             Assert.Single(services);
         }
+        [Fact]
+        public async Task CreateServiceAsync_FaultyNetwork_Throws()
+        {
+            await Assert.ThrowsAsync<DockerApiException>(() => _client.Swarm.CreateServiceAsync(new ServiceCreateParameters()
+            {
+
```
jterry75: NIT: extra newline

test/Docker.DotNet.Tests/QueryStringTests.cs
```diff
@@ -9,18 +9,18 @@ public class QueryStringTests
         [Fact]
```
jterry75: Can we add a test case for

`var p = new ServicesListParameters { Filters = new ServiceFilter() };`

Want to make sure we handle the null enumerable correctly.

src/Docker.DotNet/Endpoints/SwarmOperations.cs
dud380: I added a new exeption, dockerswarmexception which takes the deserialized message when created. So it's more typed style now, and I think it makes more sence now to have custom handler.

src/Docker.DotNet/Models/ServiceFilter.cs
dud380: It is null as default now

test/Docker.DotNet.Tests/ISwarmOperationsTests.cs
dud380: Fixed

test/Docker.DotNet.Tests/ISwarmOperationsTests.cs
dud380: Fixed

test/Docker.DotNet.Tests/QueryStringTests.cs
dud380: New test added

src/Docker.DotNet/Endpoints/SwarmOperations.cs
dud380: The http status codes returned from docker api have different meaning, and a json is returned from docker api with a message, like "node is not part of a swarm".

src/Docker.DotNet/Endpoints/SwarmOperations.cs
```diff
@@ -15,7 +15,8 @@ internal class SwarmOperations : ISwarmOperations
         {
             if (statusCode < HttpStatusCode.OK || statusCode >= HttpStatusCode.BadRequest)
             {
-                throw new DockerApiException(statusCode, responseBody);
+                var deserializedBody = new JsonSerializer().DeserializeObject<JSONError>(responseBody);
```
jterry75: Use `this._client.JsonSerializer` instead of `new JsonSerializer()` so that the user settings apply to this as well.

src/Docker.DotNet/Models/ServiceFilter.cs
```diff
@@ -26,19 +26,12 @@ public ServiceCreationMode[] Mode
         {
             get
             {
-                if (this["mode"].Length == 0)
```
jterry75: I think we have to be more careful here:
```C#
get
{
    var mode = this["mode"]
    if mode == null
    {
        return null
    }
    var lst = new ServiceCreationMode[mode.Length];
    // ... rest of your code
}
```
```C#
// I think the ?. will work here but if it doesn't add the null assignment check.
set => this["mode"] = value?.Select(m => m.ToString()).ToArray();
```

src/Docker.DotNet/Endpoints/SwarmOperations.cs
dud380: It's not accessible from the delegate, unless we change the ApiResponseErrorHandlingDelegate to take jsonserializer as a parameter, which will affect other handlers. It would make sense since all exceptions from docker api have json in the response body. Maybe it's something to consider in future improvments?

src/Docker.DotNet/Models/ServiceFilter.cs
dud380: Fixed

src/Docker.DotNet/Endpoints/SwarmOperations.cs
galvesribeiro: It could be made non-static, I think...

dud380: As far I know it cannot be made non-static, but I might be wrong?

galvesribeiro: What is wrong with this?

image

dud380: The problem is that you can't access "this", because it's an anonymous method. Try initialize swarmresponsehandler with a non-static method in Test, and you will get compilation error.

dud380: My mistake, it's working now. Actually I got the "A field initializer cannot reference the nonstatic field, method, or property" and thought it had to do with the non-static method, and it sort of had, but the problem was that the delegate had to be instanciated in the constructor.

Thanks for your patience.

jterry75:
@galvesribeiro - PTAL

PR #364, opened 2018-12-12
progress can be null in CreateImageAsync

Raffa50:
Optional progress can be very useful (instead of passing an empty one: new Progress<JSONMessage>( JSONMessage _ => {}) )


src/Docker.DotNet/Endpoints/StreamUtil.cs
```diff
@@ -45,7 +45,8 @@ internal static async Task MonitorStreamForMessagesAsync<T>(Task<Stream> streamT
                                 var prog = client.JsonSerializer.DeserializeObject<T>(line);
                                 if (prog == null) continue;
 
-                                progress.Report(prog);
+                                if(progress!= null)
```
jterry75: Rather than this we should just have the call to monitor split so we don't call monitor if we don't have a progress.

PR #366, opened 2019-01-29
Make ExecOperations available

glucaci:
Do you have any reason not to make ExecOperations available through DockerClient?

Thanks!


jterry75:
@galvesribeiro - PTAL.

PR #376, opened 2019-03-08
Synchronize write to a buffer in BufferedReadStream

arkrasik:


src/Docker.DotNet/Microsoft.Net.Http.Client/BufferedReadStream.cs
```diff
@@ -159,7 +164,25 @@ private async Task EnsureBufferedAsync(CancellationToken cancel)
...
+                try
+                {
+                    if (validBuffer)
+                    {
+                        _bufferCount = await _inner.ReadAsync(_buffer, _bufferOffset, _buffer.Length, cancel).ConfigureAwait(false);
+                    }
+                }
+                finally
+                {
+                    if ((Interlocked.Decrement(ref _bufferRefCount) == 0) && validBuffer)
```
jterry75: Shouldn't this be `if (validBuffer && ...)` so that if the caller didn't actually increment we don' try the decrement.

arkrasik: No, since we've already incremented it:
`                bool validBuffer = Interlocked.Increment(ref _bufferRefCount) > 1;`

jterry75: Oh your right I totally misread that.

jterry75:
@galvesribeiro - Want to put a 2nd set of eyes on this?

galvesribeiro:
I've followed the thread while you where reviewing.

@arkrasik please sing the CLA and we are good 😄

jterry75:
Thanks again

PR #390, opened 2019-05-10
Added peekable streams to avoid blocking read calls on pipes with no data

novinc:
Named pipes are not able to support concurrent read/writes from separate threads. This causes code that does this to get into a deadlock when you call read on an empty pipe because the read blocks and waits for data to come in and no write operations can go through while this is happening which means the read will never see any data. There is a windows system call to peek a pipe that returns immediately even when no data is present. This allows you to be able to peek before calling read so that you can be sure data is present before calling read.


tommitytom:
Is this likely to be merged in any time soon?

jterry75:
@galvesribeiro - Do you have interest in trying to prune the backlog now that we can take some of these? This one seemed good tome many moons ago :)

galvesribeiro:
Yeah, I guess before we release the package with the current codebase we need to have a task force to review the outstanding PRs and merge/close them. 

You know this codebase better then myself so if you see something is good like this, feel free to merge directly. I'll take a look over the week in case you don't have time.

Issue #375, opened 2019-03-01
Response stream corruption when a request is cancelled


anbencic-msft:
**Output of `dotnet --info`:**

```
.NET Core SDK (reflecting any global.json):
 Version:   2.1.503
 Commit:    4c506e0f35

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.17763
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.1.503\

Host (useful for support):
  Version: 2.1.7
  Commit:  cca5d72d48

.NET Core SDKs installed:
  2.1.201 [C:\Program Files\dotnet\sdk]
  2.1.202 [C:\Program Files\dotnet\sdk]
  2.1.402 [C:\Program Files\dotnet\sdk]
  2.1.503 [C:\Program Files\dotnet\sdk]

.NET Core runtimes installed:
  Microsoft.AspNetCore.All 2.1.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.All 2.1.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
  Microsoft.AspNetCore.App 2.1.4 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 2.1.7 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 2.0.7 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.0.9 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.4 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 2.1.7 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
```

**What version of Docker.DotNet?:**

```
// Docker.DotNet, Version=3.125.2.0, Culture=neutral, PublicKeyToken=null
// Global type: <Module>
// Architecture: AnyCPU (64-bit preferred)
// Runtime: .NET 4.0
```

**Steps to reproduce the issue:**

```
using System;
using System.Diagnostics;
using System.Threading;
using Docker.DotNet.Models;
using Newtonsoft.Json;

namespace ConsoleApp13
{
[...]

PR #394, opened 2019-05-16
#375 Response stream corruption when a request is cancelled

Duracell1989:
I've ran into the same issue as #375; this code works in my project.
No more `IOException`s after cancelling the token.

### DIFF
--- a/src/Docker.DotNet/Endpoints/StreamUtil.cs
+++ b/src/Docker.DotNet/Endpoints/StreamUtil.cs
@@ -1,10 +1,10 @@
+using Newtonsoft.Json;
 using System;
 using System.IO;
 using System.Net.Http;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-using Newtonsoft.Json;
 
 namespace Docker.DotNet.Models
 {
@@ -17,15 +17,12 @@ namespace Docker.DotNet.Models
             using (var stream = await streamTask)
             {
                 // ReadLineAsync must be cancelled by closing the whole stream.
-                using (cancel.Register(() => stream.Dispose()))
+                using (var reader = new StreamReader(stream, new UTF8Encoding(false)))
                 {
-                    using (var reader = new StreamReader(stream, new UTF8Encoding(false)))
+                    string line;
+                    while ((line = await reader.ReadLineAsync()) != null && !cancel.IsCancellationRequested)
                     {
-                        string line;
-                        while ((line = await reader.ReadLineAsync()) != null)
-                        {
-                            progress.Report(line);
-                        }
+                        progress.Report(line);
                     }
                 }
             }
@@ -49,31 +46,28 @@ namespace Docker.DotNet.Models
         {
             using (var response = await responseTask)
             {
-                await client.HandleIfErrorResponseAsync(response.StatusCode, response);
-
                 using (var stream = await response.Content.ReadAsStreamAsync())
                 {
                     // ReadLineAsync must be cancelled by closing the whole stream.
-                    using (cancel.Register(() => stream.Dispose()))
+                    using (var reader = new StreamReader(stream, new UTF8Encoding(false)))
                     {
-                        using (var reader = new StreamReader(stream, new UTF8Encoding(false)))
+                        string line;
+                        try
                         {
-                            string line;
-                            try
+                            while ((line = await reader.ReadLineAsync()) != null && !cancel.IsCancellationRequested)
                             {
-                                while ((line = await reader.ReadLineAsync()) != null)
-                                {
-                                    var prog = client.JsonSerializer.DeserializeObject<T>(line);
-                                    if (prog == null) continue;
+                                await client.HandleIfErrorResponseAsync(response.StatusCode, response);
 
-                                    progress.Report(prog);
-                                }
+                                var prog = client.JsonSerializer.DeserializeObject<T>(line);
+                                if (prog == null) continue;
+
+                                progress.Report(prog);
                             }
-                            catch (ObjectDisposedException)
-                            {
-                                // The subsequent call to reader.ReadLineAsync() after cancellation
-                                // will fail because we disposed the stream. Just ignore here.
-                            }
+                        }
+                        catch (ObjectDisposedException)
+                        {
+                            // The subsequent call to reader.ReadLineAsync() after cancellation
+                            // will fail because we disposed the stream. Just ignore here.
                         }
                     }
                 }
--- a/test/Docker.DotNet.Tests/ISystemOperations.Tests.cs
+++ b/test/Docker.DotNet.Tests/ISystemOperations.Tests.cs
@@ -111,10 +111,11 @@ namespace Docker.DotNet.Tests
 
             await _client.Images.CreateImageAsync(new ImagesCreateParameters { FromImage = "hello-world" }, null, progressJSONMessage);
 
-            var task = Task.Run(() => _client.System.MonitorEventsAsync(new ContainerEventsParameters(), progressMessage, cts.Token));
+            var task = _client.System.MonitorEventsAsync(new ContainerEventsParameters(), progressMessage, cts.Token);
 
             await _client.Images.TagImageAsync(repository, new ImageTagParameters { RepositoryName = repository, Tag = newTag });
 
+            Thread.Sleep(500); // Let's wait some time before cancelling the monitor stream to make sure we got some progress back.
             cts.Cancel();
 
             bool taskIsCancelled = false;


Duracell1989:
I want to use this package, but I need this fix for it to work. Can this PR be merged?

galvesribeiro:
@jterry75 any thoughts?

dgvives:
@Duracell1989 did you experience this problem of stream not having <CR/LF> and execution being trapped there?
#434

Duracell1989:
> @Duracell1989 did you experience this problem of stream not having <CR/LF> and execution being trapped there?
> #434

No, the problem was that when cancelling; it would throw `IOException`s.

Duracell1989:
@jterry75; Rebased with the latest master- thus, resolved all the merge conflicts.

jterry75:
@Duracell1989 - Super sorry I thought that #481 fixed this but now I looked back and realized we only fixed one of the three similar methods. Given the change we did there by removing the `stream.Dispose` on task cancellation we no longer should ever see the `ObjectDisposedException`. I think you want to follow the pattern for the `WithCancellation` we did in the other one on the `ReadLineAsync()` call to unblock the task right?

dgvives:
I wouldn't recommend WithCancellation extension, produces different results depending on docker response time, causing tests to fail on CI/CD pipeline tests.
I know I pushed it, didn't know better then

Get Outlook for Android<https://aka.ms/ghei36>

Duracell1989:
As suggested by @dgvives: I removed the WithCancellation and have rewritten this to the same cancellation as the rest of the code.

dgvives:
@Duracell1989 apologies, I didn't explain myself properly

`WithCancellation` extension method can not be just removed as it is required to cancel `MonitorStreamForMessagessAsync`

Commit [d7372c0](https://github.com/dotnet/Docker.DotNet/pull/394/commits/d7372c0111f3e739b9a641e81e14d77c53df511a)  didn't pass the tests, being cancelled after 360m build timeout on the CI/CD pipeline. 
[Details](https://github.com/dotnet/Docker.DotNet/pull/394/checks?check_run_id=1820702834)

That I have in mind is to replace `WithCancellation` extension by using `TaskCompletionSource` to exit loop but remove exception being thrown. 
Original idea was @suneption on #343, I just incorporated the changes required to pass CI/CD pipeline test. 


After some unexpected results,  I just updated this part avoid exceptions from uncertain execution frames.


```csharp
        internal static async Task MonitorStreamForMessagesAsync<T>(Task<Stream> streamTask, CancellationToken cancel, IProgress<T> progress)
        {

            var tcs = new TaskCompletionSource<bool>();

            using (cancel.Register(() => tcs.TrySetCanceled(cancel)))
            {
                while (await await Task.WhenAny(jsonReader.ReadAsync(default), tcs.Task))
                {
                    var ev = Serializer.Deserialize<T>(jsonReader);
                    progress.Report(ev);
                }
            }
        }
```


Haven't pushed this yet as it is pending tests. 
BTW, I checked your idea on checking Token.IsCancellationRequested and it was working on the two previously updated methods, 👍 .

Duracell1989:
@dgvives; you're going to add my `cancel.IsCancellationRequested` change to another branch of you?
Or; do you want me to make some changes to approve this issue and accept the PR?

dgvives:
@Duracell1989 I won't add your changes to the branch I'm currently working on, I had to learn it is considered bad contributor etiquette.
I suggest you rollback to [9f18f36](https://github.com/dotnet/Docker.DotNet/pull/394/commits/9f18f3690c4e30334c17348c50f15dd24e65eeab), their tests [results](https://github.com/dotnet/Docker.DotNet/runs/1805220627) were fine. 

Besides this it would be nice to include some test case that calls the modified method.
I'm working on tests cases for that piece of [code](https://github.com/dotnet/Docker.DotNet/pull/394#issuecomment-772622348) I posted before, and won't submit PR for it until tests are added.

You decide to go ahead with tests and like to see any example, I'm happy to help, just ping me.

dgvives:
@Duracell1989 by looking at your modification for `MonitorStreamForMessagesAsync` method, looks like condition `cancel.IsCancelationRequested` won't be hit as it exits earlier when `OperationCanceledException` is thrown in `.WithCancelation` extension method

Have you tested `MonitorStreamAsync` and `MonitorResponseForMessagesAsync` methods after your changes? How did you do it? 
It's being a challenge to test those

Duracell1989:
@dgvives I didn't test that specific change - also because I found it a challenge. Let me revert that one in that function.

dgvives:
@Duracell1989 I found there is a chance of avoidable exception being thrown when `MonitorResponseForMessagesAsync` task is cancelled. 

DockerApiException is being thrown when `await client.HandleIfErrorResponseAsync(response.StatusCode, response);` but because reader task is cancelled it won't read.

To solve this, move `await client.HandleIfErrorResponseAsync(response.StatusCode, response);` first line after opening bracket on while. 
This seems to solves the issue.

image

I'd like to merge your changes into the branch I'm working on because it fixes other exceptions being thrown

dgvives:
@Duracell1989 I closed #487, tests results could be different on each execution
You can check tests results for #394

Duracell1989:
@dgvives; Yeah, I know how to run tests. But if I **run all** the tests; there is this 1 tests that fail. But if you run that failing test on it's own. If passed. So, some data isn't getting cleared between multiple tests.
I hoped it was only on my machine - but it is also on the build agent. So, I'm still investigating why this is happening.

Duracell1989:
I think we have a race-condition failure:
By running multiple tests, the `_onMessageCalled` function **isn't** called (no idea why) and the `Assert.True(wasProgressCalled);` fails.
When running the `MonitorEventsAsync_Succeeds` test separate, the `_onMessageCalled` function is getting called, so here the `Assert.True(wasProgressCalled);` succeeded.

dgvives:
I've had the same problem while trying to validate issues were solved. 
Initially my intention was to fix `WaitContainerAsync` call but it turned to be more complicated than I expected. 

I ran into the same problems you faced, different test results on different executions for the same tests, tests running fine locally but failing on CI/CD pipeline, tests running ok individually but failing when all together, tests never ending and causing TimeoutException on build agent.

Duracell1989:
I've fixed this by adding the `Thread.Sleep` to the test. This way, we give the `Stream` some time do report back some progress; see [here](https://github.com/dotnet/Docker.DotNet/pull/394/files#diff-452b0ba41d47f4b334d59ae5c23c3028a4a2ca89cec379cb761ba9e2c5856b20R118)

src/Docker.DotNet/Endpoints/StreamUtil.cs
```diff
@@ -17,15 +17,12 @@ internal static async Task MonitorStreamAsync(Task<Stream> streamTask, DockerCli
             using (var stream = await streamTask)
             {
                 // ReadLineAsync must be cancelled by closing the whole stream.
-                using (cancel.Register(() => stream.Dispose()))
+                using (var reader = new StreamReader(stream, new UTF8Encoding(false)))
                 {
-                    using (var reader = new StreamReader(stream, new UTF8Encoding(false)))
+                    string line;
+                    while ((line = await reader.ReadLineAsync()) != null && !cancel.IsCancellationRequested)
```
jterry75, line 23: I'm still confused by this. How does calling cancel unblock a pending `ReadLineAsync` call? This code seems to be required to wait for the next line before it would fall out of the loop.

Duracell1989: It doesn't stop the current `ReadLineAsync`. It is a `while`-loop; it keeps reading while there is still data and will only stop when it's done reading all the data - which is never because it is a monitoring stream which keeps reporting back data from Docker. The new code `cancel.IsCancellationRequested` is add so that when the process is cancelled: it finishes the current `ReadLineAsync` and quits the `while`-loop.

jterry75:
@Duracell1989 / @dgvives - Which of these changes are we taking looks like both are still being actively worked on?

dgvives:
These changes were merged into #487 and it continued from there until tests success

jterry75:
Sounds good can we close this one?

Duracell1989:
Best practice is to merge a PR if it adds value. Waiting for another PR which could even be better isn't great, because you never know when this other PR will be approved (or never at all).

In this case; the #487 has been closed and no timeline is present when  other improvements will be finished.
So, I would suggest to merge this - to at least resolve the IO Exception and later; in another PR we could improve upon this, if that is needed.

jterry75:
@Duracell1989 / @dgvives - The other was closed are we wanting to take this one first and work out the rest later?

dgvives:
I hit a wall being unable to cancel `MonitorStreamForMessagesAsync` call with cancellation token, and I'm not sure how long it's going to take me to find a proper solution, as there are custom handlers and stream readers involved.

There is a reason behind calling Stream.Dispose(), once `JsonTextReader` starts reading, it won't stop until end of stream or full json token is read. 
This is a problem when trying to cancel `dockerClient.Container.GetContainerLogs` using `ContainerLogsParameters{ Follor = true }`, or `dockerClient.Container.GetContainerStats` using `ContainerStatsParameters{ Stream = true }`

The stream is continuous and there is no way to stop reading because the reader is using calls to `stream.Read()` method and this won't stop once finished.

I'm reviewing implementation on source code for these methods, but it's taking a while. I can't recommend removing `stream.Dispose` until a tested alternative have been found

Duracell1989:
From what I understand:
You register a anonymous function which does a `stream.Dispose()` on the `CancellationToken`. This means that the `Stream` get's dispose as soon as the cancellation is requested.
But, the problem here is that the `Stream` could still be in use (still reading) within the `while` loop and still get's accessed but it's already disposed of (because of the anonymous function registered on the `CancellationToken`.
Because the `Stream` is within a `using` statement (for example: `using (var stream = await streamTask)` and `using (var stream = await response.Content.ReadAsStreamAsync())`) it doesn't need to be disposed of immediately when the `CancellationToken` is cancelled. That's why checking the !cancel.IsCancellationRequested makes sure we exit the `while` loop (and not using the `Stream` anymore) and after that (because of the `using` statement) the `Stream` get's disposed of at the correct time, when we're stopped using the `Stream`.

jterry75:
@Duracell1989 - The `stream.Dispose()` had the intent of unblocking the outstanding read to `ReadAsStreamAsync` since no cancellation is supported don this read. Its not really about making sure to dispose the object its about releasing the wait. I dont see how the `!cancel.IsCancellationRequested` is unblocked until the next read comes in which may be some unknown amount of time in the future right?

dgvives:
Found the problem

Basically, deep inside it goes inside a `while(read())`  with no possibility to exit by using CancellatonToken
I has ended up being much more complicated than it seemed

Detailed  explanation here https://github.com/dotnet/Docker.DotNet/pull/487#issuecomment-786819772

PR #405, opened 2019-07-02
Fix SocketLeak when CreateImageAsync throws Exception

JennyLawrance:
Seeing a leak of sockets in CLOSE_WAIT state (ubuntu 16.04), .NET Core version, when a CreateImageAsync is issued to a non-existent image. Upon reviewing the code, there are multiple places where the HttpResponseMessage object is not disposed (when using stream).
Verified fix in the environment where socket leak was observed.


jterry75:
@JennyLawrance - Can you rebase these to exclude the merge commits?

jterry75:
Took #406 instead.

PR #432, opened 2020-01-19
Update model

cfauchere:
Updating model to leverage more recent docker functionality

### DIFF
--- a/src/Docker.DotNet/Models/ContainerUpdateParameters.Generated.cs
+++ b/src/Docker.DotNet/Models/ContainerUpdateParameters.Generated.cs
@@ -31,8 +31,10 @@ namespace Docker.DotNet.Models
                 this.CpusetCpus = UpdateConfig.CpusetCpus;
                 this.CpusetMems = UpdateConfig.CpusetMems;
                 this.Devices = UpdateConfig.Devices;
-                this.DiskQuota = UpdateConfig.DiskQuota;
+                this.DeviceCgroupRules = UpdateConfig.DeviceCgroupRules;
+                this.DeviceRequests = UpdateConfig.DeviceRequests;
                 this.KernelMemory = UpdateConfig.KernelMemory;
+                this.KernelMemoryTCP = UpdateConfig.KernelMemoryTCP;
                 this.MemoryReservation = UpdateConfig.MemoryReservation;
                 this.MemorySwap = UpdateConfig.MemorySwap;
                 this.MemorySwappiness = UpdateConfig.MemorySwappiness;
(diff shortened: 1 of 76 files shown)


src/Docker.DotNet/Models/ContainerUpdateParameters.Generated.cs
```diff
@@ -117,7 +125,7 @@ public ContainerUpdateParameters(UpdateConfig UpdateConfig)
         public bool? OomKillDisable { get; set; }
 
         [DataMember(Name = "PidsLimit", EmitDefaultValue = false)]
-        public long PidsLimit { get; set; }
+        public long? PidsLimit { get; set; }
```
jterry75: This is a breaking change right?

src/Docker.DotNet/Models/ContainerUpdateParameters.Generated.cs
```diff
@@ -98,12 +100,18 @@ public ContainerUpdateParameters(UpdateConfig UpdateConfig)
         [DataMember(Name = "Devices", EmitDefaultValue = false)]
         public IList<DeviceMapping> Devices { get; set; }
 
-        [DataMember(Name = "DiskQuota", EmitDefaultValue = false)]
-        public long DiskQuota { get; set; }
```
jterry75: This is a breaking change right? What replaced it?

jterry75:
Just a few questions but this is mostly additive so I think its great!

src/Docker.DotNet/Models/ContainerUpdateParameters.Generated.cs
cfauchere: Looks like it is to me. Unfortunately, I do not see what options we have since this reflects changes made in the docker API.

src/Docker.DotNet/Models/ContainerUpdateParameters.Generated.cs
cfauchere: I am not sure. I looked at the docker API documentation and did not find anything called DiskQuota. 
(https://docs.docker.com/engine/api/v1.40/#operation/ContainerUpdate)
I also looked in the [repo](https://github.com/moby/moby/blob/b95fad8e51bd064be4f4e58a996924f343846c85/api/types/container/host_config.go) and did not find anything obvious.
The net is I do not even know where this field came from in the first place

jterry75: Super weird. This seems to have been removed in `v1.40`. But I cant see any notes on why. Last model it was in is: `https://docs.docker.com/engine/api/v1.39/#operation/ContainerUpdate`

src/Docker.DotNet/Models/ContainerUpdateParameters.Generated.cs
jterry75: Ok thats fine. It needs to be null anyways otherwise its default of 0 will not actually set to unlimited as it will not be serialized in the first place.

src/Docker.DotNet/Models/ContainerUpdateParameters.Generated.cs
cfauchere: Yes, looks like it was remove moving to 1.40. I do not think it's a breaking change because likely the field will be ignored by docker. It might be perceived by users as regression though since the field is moot. At the end of the day I do not think we can do better than what docker api does.

jterry75: Agreed.

Crozin:
Can we expect a NuGet release anytime soon?

jterry75:
> Can we expect a NuGet release anytime soon?

Sorry not at this time. It would be best for you to build/sign/deploy this as your application sees fit for your use. Thanks and sorry for any inconvenience.

PR #454, opened 2020-07-11
Sign build

galvesribeiro:
This PR make the following changes to allow new releases to come from this repo:

1. Drop AppVeyor and use GitHub Actions workflows;
2. Deterministic Build and versioning using [NBGV](https://github.com/dotnet/Nerdbank.GitVersioning);
3. Code sign using [.Net Foundation Sign Service](https://github.com/dotnet/SignService);
4. Merge all common props into a props file;
5. Target `netstandard2.0`

Note that this PR is not the release yet. It is just adding the assets/changes required to a new one. Once this is merged, we will review/merge the outstanding PRs and make a proper release.

@jterry75 would love to have your eyes on this to see if anything breaks.

@clairernovotny could you please check if the sign process is appropriated? I haven't found any other repos or samples using GitHub Actions, so all my tests where done locally on my Mac.

Thanks!

completes: https://github.com/dotnet-foundation/projects/issues/67


clairernovotny:
Seems ok

.gitignore
```diff
@@ -2,7 +2,6 @@
 ## files generated by popular Visual Studio add-ons.
 
 .vs
-.vscode
```
jterry75, line 5: Can we leave this?

galvesribeiro: @jterry75 I added that because unlike the `.vs`, it only contains the "build config" when you are using VSCode. It is just a couple of JSON files and are not machine specific. The `.vs` one has the extension index database which should be excluded.

jterry75: No if you have made any project specific settings/debug/etc it will be in .vscode. Its a pretty standard practice to .gitignore the .vscode folder

galvesribeiro: Ok ok, I usually have it as a standard config to build/debug the project anywhere without machine or OS specific things. And all other project I've worked actually keep that on the repo so it is simples to build/debug. Think on that as a simple `build.sh`/ `run.sh`.

Bot ok, I understand your point. Will put that back when I get back to a PC and merge this.

Thanks!

tfenster:
@galvesribeiro @clairernovotny as I had to once again build this library manually to get fix in one of my dependent projects, I would like to politely ask when there is a chance for more progress?

galvesribeiro:
Hey folks, I've merge and tried to release the package Today. However, even after nuget successfully pushed the package, it fail validating the Signature and NuGet sent me an email about the certificate thumbprint not matching:

image

@clairernovotny can you give a hand here? What I'm probably missing?

clairernovotny:
Can you try pushing again? It should let you push the same package version since it failed validation.

galvesribeiro:
Yay! 

image

Now it is fixed :) Thanks @clairernovotny

PR #455, opened 2020-07-15
Update README.md

gittadesushil:
I am using Docker.DotNet latest stable version 3.125.2 but Docker client initialization is giving error to me DockerClient client = new DockerClientConfiguration()
Can you please verify? May be I am wrong!

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,9 +54,10 @@ or to connect to your local [Docker for Windows](https://docs.docker.com/docker-
 
 ```csharp
 using Docker.DotNet;
-DockerClient client = new DockerClientConfiguration()
+DockerClient client = new DockerClientConfiguration(new Uri("tcp://localhost:2375"))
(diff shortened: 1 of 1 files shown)


dgvives:
Hello, 
I think you no longer require to expose docker http port to connect locally.  I suggest take a look how this guy does:

https://github.com/Deffiss/testenvironment-docker/blob/630f20f9c2b5f5cc8f82b92932e1ea5390ea62a3/src/TestEnvironment.Docker/DockerEnvironmentBuilder.cs#L138

Hope this helps you

Emdot:
Docker for Windows allows you to connect this way, but it's disabled by default for security reasons, so it probably shouldn't be part of the normal instructions. Fortunately, this isn't necessary, because Docker.DotNet can correctly connect to Docker for Windows over IPC by default.

PR #460, opened 2020-08-25
README.md update from [Obsolete] MonitorEventsAsync example to latest

trendfollowerpl:
Hello.
Section : https://github.com/dotnet/Docker.DotNet#example-dealing-with-stream-responses
contains mix of obsolete return type and a new MonitorEventsAsync method signature.
This PR fixes that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,8 +150,10 @@ To obtain this stream you can use:
 
 ```csharp
 CancellationTokenSource cancellation = new CancellationTokenSource();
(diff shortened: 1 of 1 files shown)


tinohager:
I think this PR is already outdated again.

This example works for me
```
var progress = new Progress<Message>(o => Console.WriteLine($"{o.Action} {o.Actor.ID} {o.From} {o.ID} {o.Scope} {o.Status} {o.Type}"));

await client.System.MonitorEventsAsync(new ContainerEventsParameters(), progress);
```

PR #472, opened 2020-10-14
Updated specgen tool to use a go module and added get service logs api call

The-TT-Hacker:
Updated the specgen tool so it uses the newer go modules instead of godep.
I also updated the docker version imported to the latest 19.03.13.

This was tested on a fresh install of go on a new machine. All you have to do to update the models is install go, update the docker repo which is imported using go get -u github.com/docker/docker@<release-tag-or-hash> and then run the update-generate-code.cmd script. It downloads all of the correct dependencies, builds the program and generates the models. This has been updated in the readme for specgen.

I have also added a call to the get service logs endpoint. To do this I had to add a model which was the reason for updating the go specgen program.


ACoderLife:
This pull request can be closed when [@589](https://github.com/dotnet/Docker.DotNet/pull/589) is in.
The important backend changes were done in https://github.com/dotnet/Docker.DotNet/pull/480

PR #473, opened 2020-10-15
Added MultiplexedStreamReader

The-TT-Hacker:
Added a simple reader for the multiplexed stream used for getting logs from containers which can read lines as strings. Also works with the service logs pull request i've made.
Ideally the multiplex stream would extend the stream class or something so we can use the normal stream reader but this is working for me.

### DIFF
--- a/src/Docker.DotNet/MultiplexedStreamReader.cs
+++ b/src/Docker.DotNet/MultiplexedStreamReader.cs
@@ -14,7 +14,7 @@ namespace Docker.DotNet
             _stream = stream;
         }
 
-        public async Task<string> ReadLineAsync(CancellationToken token)
+        public async Task<string> ReadLineAsync(CancellationToken token = default)
         {
             var line = new List<byte>();
 
--- a/src/Docker.DotNet/MultiplexedStreamReader.cs
+++ b/src/Docker.DotNet/MultiplexedStreamReader.cs
@@ -1,0 +1,44 @@
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Docker.DotNet
+{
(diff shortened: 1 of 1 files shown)


src/Docker.DotNet/MultiplexedStreamReader.cs
```diff
@@ -0,0 +1,44 @@
...
+        {
+            _stream = stream;
+        }
+
+        public async Task<string> ReadLineAsync(CancellationToken token = default)
+        {
+            var line = new List<byte>();
+
+            var buffer = new byte[1];
+
```
dgvives, line 22: @The-TT-Hacker I tested this method trying to get rid of headers being shown on `MonitorStreamAsync` method and haven't managed to cancel it. 
Would you confirm if expected behaviour is to read until 0 bytes are returned or to stop reading when no more lines are retrieved?
When `res` equals 0, it loops indefinitelly.
This is test I used
https://gist.github.com/dgvives/2620960049191797ea0d871c339d9bc2#file-program-cs-L1 

It would be nice if MultiplexedStreamReader could detect stream being multiplexed  and provide output por ReadLineAsync without control chars.

Have you checked this running containers on different platforms? Because of EOL

ACoderLife:
@galvesribeiro I have updated this one with cancellation support.

src/Docker.DotNet/MultiplexedStreamReader.cs
galvesribeiro: @ACoderLife hey! Sorry for taking long. I'm cleaning up and merging so PRs with fixes and improvements before I start the wave of changes on the whole codebase. 

That being said, have you checked @dgvives concern with the infinite loop?

Thank you!

Issue #434, opened 2020-01-31
Events monitoring stream does not seems to always include CR or LF character at the end of every messages


simardst:
**Output of `dotnet --info`:**
```
.NET Core SDK (reflecting any global.json):
 Version:   3.1.101
 Commit:    b377529961

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.18362
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.1.101\

Host (useful for support):
  Version: 3.1.1
  Commit:  a1388f194c
```

**What version of Docker.DotNet?:**
```
3.125.2
```

**Docker Version:**
```
Client: Docker Engine - Community
 Version:           19.03.5
 API version:       1.40
 Go version:        go1.12.12
 Git commit:        633a0ea
 Built:             Wed Nov 13 07:22:37 2019
 OS/Arch:           windows/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.5
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.12
  Git commit:       633a0ea
  Built:            Wed Nov 13 07:29:19 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.10
  GitCommit:        b34a5c8af56e510852c35414db4c1f4fa6172339
 runc:
  Version:          1.0.0-rc8+dev
  GitCommit:        3e425f80a8c931f88e6d94a8c831b9d5aa481657
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683
```

**Docker Info:**
```
Client:                                                                                                   
 Debug Mode: false                                                                                        
                                                                                                          
Server:                                                                                                   
 Containers: 1                                                                                            
  Running: 1                                                                                              
  Paused: 0                                                                                               
  Stopped: 0                                                                                              
 Images: 10                                                                                               
 Server Version: 19.03.5                                                                                  
 Storage Driver: overlay2                                                                                 
  Backing Filesystem: extfs                                                                               
  Supports d_type: true                                                                                   
  Native Overlay Diff: true                                                                               
 Logging Driver: json-file                                                                                
 Cgroup Driver: cgroupfs                                                                                  
 Plugins:                                                                                                 
  Volume: local                                                                                           
  Network: bridge host ipvlan macvlan null overlay                                                        
[...]

PR #481, opened 2021-01-11
Minimum required changes for working tests

dgvives:
Minimum required changes for working tests.
Updated models and broken functionality


src/Docker.DotNet.BasicAuth/Docker.DotNet.BasicAuth.csproj
```diff
@@ -7,4 +7,10 @@
   <ItemGroup>
     <ProjectReference Include="..\Docker.DotNet\Docker.DotNet.csproj" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Update="Nerdbank.GitVersioning" Version="3.3.37">
```
jterry75, line 11: What is this for? We shouldnt need this

src/Docker.DotNet/DockerClient.cs
```diff
@@ -121,6 +121,7 @@ internal DockerClient(DockerClientConfiguration configuration, Version requested
             _endpointBaseUri = uri;
 
             _client = new HttpClient(Configuration.Credentials.GetHandler(handler), true);
+            _client.DefaultRequestHeaders.Add(nameof(_endpointBaseUri.Host), _endpointBaseUri.Host);
```
jterry75, line 124: Was this a bug? Why do we need it?

src/Docker.DotNet/DockerClientConfiguration.cs
```diff
@@ -16,8 +16,14 @@ public class DockerClientConfiguration : IDisposable
 
         private static Uri LocalDockerUri()
         {
-            var isWindows = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
-            return isWindows ? new Uri("npipe://./pipe/docker_engine") : new Uri("unix:/var/run/docker.sock");
+            var dockerHostVar = Environment.GetEnvironmentVariable("DOCKER_HOST");
```
jterry75, line 19: This is new support for an env in the command window right?

src/Docker.DotNet/Endpoints/StreamUtil.cs
```diff
@@ -30,32 +31,47 @@ internal static async Task MonitorStreamAsync(Task<Stream> streamTask, DockerCli
...
+            try
             {
-                // ReadLineAsync must be cancelled by closing the whole stream.
+                using (var stream = await streamTask)
                 using (cancel.Register(() => stream.Dispose()))
+                using (var reader = new StreamReader(stream, new UTF8Encoding(false)))
+                using (var jsonReader = new JsonTextReader(reader) { SupportMultipleContent = true })
                 {
-                    using (var reader = new StreamReader(stream, new UTF8Encoding(false)))
+                    while (jsonReader.Read())
```
jterry75, line 42: If we used ReadAsync on this then we wouldnt need to do the stream.Dispose above on cancellation and then we wouldnt need this complex exception processing anyways right?

src/Docker.DotNet/Models/ContainerExecInspectResponse.Generated.cs
```diff
@@ -5,7 +5,7 @@ namespace Docker.DotNet.Models
     [DataContract]
     public class ContainerExecInspectResponse // (types.ContainerExecInspect)
     {
-        [DataMember(Name = "ExecID", EmitDefaultValue = false)]
+        [DataMember(Name = "ID", EmitDefaultValue = false)]
```
jterry75, line 8: Did we always have this wrong of did Docker change this?

src/Docker.DotNet/Models/ContainerSpec.Generated.cs
```diff
@@ -78,7 +78,13 @@ public class ContainerSpec // (swarm.ContainerSpec)
         [DataMember(Name = "Sysctls", EmitDefaultValue = false)]
         public IDictionary<string, string> Sysctls { get; set; }
 
-        [DataMember(Name = "Capabilities", EmitDefaultValue = false)]
-        public IList<string> Capabilities { get; set; }
+        [DataMember(Name = "CapabilityAdd", EmitDefaultValue = false)]
```
jterry75, line 81: This must have been a Docker breaking change?

src/Docker.DotNet/Models/GlobalJob.Generated.cs
```diff
@@ -0,0 +1,9 @@
+using System.Runtime.Serialization;
+
+namespace Docker.DotNet.Models
+{
+    [DataContract]
+    public class GlobalJob // (swarm.GlobalJob)
+    {
```
jterry75, line 7: Odd this has no members?

src/Docker.DotNet/Models/Limit.Generated.cs
```diff
@@ -0,0 +1,17 @@
+using System.Runtime.Serialization;
+
+namespace Docker.DotNet.Models
+{
+    [DataContract]
+    public class Limit // (swarm.Limit)
```
jterry75, line 6: All of the Swarm stuff we have called "SwarmLimit", "SwarmJobStatus" perviously since these are not namespaced to avoid name collisions

src/Docker.DotNet/Models/ServiceFilter.cs
```diff
@@ -9,30 +9,30 @@ public class ServicesListParameters
         [QueryStringParameter("filters", false, typeof(MapQueryStringConverter))]
         public ServiceFilter Filters { get; set; }
     }
-
-    public class ServiceFilter : Dictionary<string, string>
+
```
jterry75, line 12: NIT: extra space

src/Docker.DotNet/Models/ServiceFilter.cs
```diff
@@ -9,30 +9,30 @@ public class ServicesListParameters
         [QueryStringParameter("filters", false, typeof(MapQueryStringConverter))]
         public ServiceFilter Filters { get; set; }
     }
-
-    public class ServiceFilter : Dictionary<string, string>
+    
+    public class ServiceFilter : Dictionary<string, string[]>
```
jterry75, line 13: This is a breaking change. Did docker break it?

src/Docker.DotNet/Models/TLSInfo.Generated.cs
```diff
@@ -10,9 +9,9 @@ public class TLSInfo // (swarm.TLSInfo)
         public string TrustRoot { get; set; }
 
         [DataMember(Name = "CertIssuerSubject", EmitDefaultValue = false)]
-        public IList<byte> CertIssuerSubject { get; set; }
+        public string CertIssuerSubject { get; set; }
```
jterry75, line 12: Hmm? Are you sure you can pass a cert this way? Seems wrong

tools/specgen/modeldefs.go
```diff
@@ -205,9 +205,8 @@ type ImagesCreateParameters struct {
 
 // ImagesListParameters for GET /images/json
 type ImagesListParameters struct {
-	MatchName string `rest:"query,filter"`
-	All       bool   `rest:"query"`
-	Filters   Args   `rest:"query"`
+	All     bool `rest:"query"`
+	Filters Args `rest:"query"`
```
jterry75, line 209: Looks like this now has `digests` too

jterry75 (changes requested):
Some questions

src/Docker.DotNet/DockerClientConfiguration.cs
dgvives: I checked online the origin for this and haven't found a suitable answer. Reverting in next commit

src/Docker.DotNet/DockerClient.cs
dgvives: At some point test were failing because there was no Host header within request. it seems to work fine now. Reverting in next commit

tools/specgen/modeldefs.go
dgvives: This struct mimics ImageListOptions https://github.com/moby/moby/blob/327daef5c19209898f11380a7bd02509d6675bcb/api/types/client.go#L239 

I've tried to specify value for digest=true or digest=false when quering docker api and it makes no difference. RepoDigests are shown by default

src/Docker.DotNet/Models/TLSInfo.Generated.cs
dgvives: Tests fail when setting this to IList<byte> and haven't found any workarround

src/Docker.DotNet/Models/ContainerExecInspectResponse.Generated.cs
dgvives: It looks like this was updated February 2020
https://github.com/moby/moby/commit/cc993a9cbf40c1a94b5bada84f1d759b72df5c4e

src/Docker.DotNet/Models/Limit.Generated.cs
dgvives: Should be renamed in next commit

src/Docker.DotNet/Models/ContainerSpec.Generated.cs
dgvives: Indeed 
https://github.com/moby/moby/commit/24f173a003727611aa482a55b812e0e39c67be65#diff-fe6d8227bdd6e327416738e82622b2b968a41c5451babe5588051df33f189c07

src/Docker.DotNet/Models/GlobalJob.Generated.cs
dgvives: https://godoc.org/github.com/docker/docker/api/types/swarm#GlobalJob

> GlobalJob is the type of a Service which executes a Task on every Node matching the Service's placement constraints. These tasks run to completion and then exit.
This type is deliberately empty.

src/Docker.DotNet/Models/ServiceFilter.cs
dgvives: Haven't found the reference, but it fails when left unmodified
```
 Docker.DotNet.DockerApiException : Docker API responded with status code=BadRequest, response={"message":"json: cannot unmarshal string into Go value of type map[string]bool"}
```

src/Docker.DotNet.BasicAuth/Docker.DotNet.BasicAuth.csproj
dgvives: Found this reference to this package
https://github.com/dotnet/Docker.DotNet/commit/bbe5d8f905568436595e3b0df5537657e35171cd#diff-5c1900dca99df01c3ddc3a884282164b04deee76e79b176e2d34cfa47f3b8907

I can't build current master branch  without adding these references. I do not know what are these for

src/Docker.DotNet/Endpoints/StreamUtil.cs
dgvives: Haven't found any other way to make events work, events endpoint is failing a for now.
Will try using ReadAsync

dgvives: Could you show me a sample of what you mean? 
This solution is not mine but @simardst
https://github.com/dotnet/Docker.DotNet/issues/434#issue-558268469
I just extended exception catchs as I was running tests

src/Docker.DotNet/Models/ServiceFilter.cs
dgvives: Haven't managed to get rid of extra space using VSCode

src/Docker.DotNet.BasicAuth/Docker.DotNet.BasicAuth.csproj
jterry75: Oh wow. I dont know how this went in. @galvesribeiro - Did you add this? We should remove this dependency IMO but not this PR's fault.

src/Docker.DotNet/Endpoints/StreamUtil.cs
jterry75: So I think that since you switched to the JSON.Net reader we can instead:

```c#
{
    bool done = false;
    do
    {
        done = await jsonReader.ReadAsync(cancel);
        // deserialize, event progress, etc.
    } while (!done)
}
```

jterry75: By doing it this way JSON.Net will return from the reader when cancellation is requested which previously was the issue that "ReadLineAsync" would not because it did not have the cancellation token. We had to close the reader out from under it in this line:
```c#
using (cancel.Register(() => stream.Dispose()))
```
Which we should no longer need

src/Docker.DotNet/Models/ContainerExecInspectResponse.Generated.cs
jterry75: Nice thanks.

src/Docker.DotNet/Models/ContainerSpec.Generated.cs
jterry75: Thanks

src/Docker.DotNet/Models/GlobalJob.Generated.cs
jterry75: Lol. Thanks

src/Docker.DotNet/Models/ServiceFilter.cs
jterry75: Yep its now a `map[string][]string` your right

src/Docker.DotNet/Models/TLSInfo.Generated.cs
jterry75: Ok np. Does appear to be a string in the json object

tools/specgen/modeldefs.go
jterry75: Ok np

jterry75 (changes requested):
I think just one question if we can improve the StreamUtil even better here otherwise this looks really good thanks for the contributions

dgvives:
@jterry75 I'm about to try your suggestions for StreamUtil, you should expect an update in next couple of hours if I succeeded on that
Thanks for your time and attention

dgvives:
@jterry75 I tried your suggestion for StreamUtil and found that `ReadAsync(token)` wasn't getting any line when token was cancelled.
Approach from #343 fits nicely, is working and validated by running tests.

The-TT-Hacker:
This pull request has my minimal changes to update the go code generation to the newer go modules.
Well I guess it's not 100% minimal because it has added the service logs endpoint which was the motivation for updating the code generation.
https://github.com/dotnet/Docker.DotNet/pull/472

src/Docker.DotNet/Models/TLSInfo.Generated.cs
dgvives: It was done in the go code generator.

dgvives: It was updated via the go code generator. My intention is to provide the minimum set of changes for tests to work, and continue adding improvements and collaborations from there.

@The-TT-Hacker I agree it would be a better solution but a minimum working base is required for any contribution to pass the CI/CD tests, get reviewed, and go ahead.

The-TT-Hacker: I dont think it's a good idea to be overriding the model types and then having to make a breaking change to the client library to revert to the correct solution. The proper fix is very minimal and not hard to include.

src/Docker.DotNet/Models/SecretSpec.Generated.cs
```diff
@@ -26,7 +26,7 @@ public SecretSpec(Annotations Annotations)
         public IDictionary<string, string> Labels { get; set; }
 
         [DataMember(Name = "Data", EmitDefaultValue = false)]
-        public IList<byte> Data { get; set; }
+        public string Data { get; set; }
```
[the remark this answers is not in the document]
dgvives: Will have a look at it later, thanks @the-tt-hacker

src/Docker.DotNet/Models/TLSInfo.Generated.cs
dgvives: Done

jterry75:
I'm ok with this at the moment. @galvesribeiro - A second look would be appreciated if you have the time

galvesribeiro:
Sure! Will do it Today. Thanks!

Get Outlook for iOS<https://aka.ms/o0ukef>

src/Docker.DotNet.BasicAuth/Docker.DotNet.BasicAuth.csproj
galvesribeiro: This can't be removed. 

It is used to stamp the assemblies with a deterministic version based on the git commit. It is only of the requirements to have the assembly signed by the sign tools which is a requirement to be on .Net Foundation. 

@dgvives that was the PR I added the sign tools.

src/Docker.DotNet.X509/Docker.DotNet.X509.csproj
```diff
@@ -8,4 +8,10 @@
   <ItemGroup>
     <ProjectReference Include="..\Docker.DotNet\Docker.DotNet.csproj" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Update="Nerdbank.GitVersioning" Version="3.3.37">
```
galvesribeiro, line 12: Please remove this. It is added by the .props file to all projects.

src/Docker.DotNet/Docker.DotNet.csproj
```diff
@@ -8,4 +8,10 @@
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Include="System.Buffers" Version="4.5.1" />
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Update="Nerdbank.GitVersioning" Version="3.3.37">
```
galvesribeiro, line 12: Same

src/Docker.DotNet/Endpoints/StreamUtil.cs
```diff
@@ -30,30 +31,16 @@ internal static async Task MonitorStreamAsync(Task<Stream> streamTask, DockerCli
 
         internal static async Task MonitorStreamForMessagesAsync<T>(Task<Stream> streamTask, DockerClient client, CancellationToken cancel, IProgress<T> progress)
         {
+            var serializer = new Newtonsoft.Json.JsonSerializer();
```
galvesribeiro, line 34: Can't we avoid Json.Net here and use System.Text.Json? 

I know this PR is only to get the tests back again. It is ok to keep it and we can deal with it on my PR later on for the next version.

However, I wouldn't create a new instance of it on every call of this method but instead, keep a shared instance (static) on this class.

src/Docker.DotNet/JsonBase64Converter.cs
```diff
@@ -0,0 +1,30 @@
...
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, Newtonsoft.Json.JsonSerializer serializer)
+        {
+            var strVal = reader.Value as string;
+
+            return Convert.FromBase64String(strVal);
+        }
+
+        public override bool CanConvert(Type objectType)
+        {
+            return objectType == typeof(IList<byte>);
```
galvesribeiro, line 27: Cache this `typeof(IList<byte>)` in a `static readonly` field and use it here to avoid reflection on every call.

src/Docker.DotNet/Models/ContainerExecInspectResponse.Generated.cs
galvesribeiro, line 8: Out of curiosity, why are we using `DataMember` instead of `JsonProperty` on this case?

test/Docker.DotNet.Tests/ISystemOperations.Tests.cs
```diff
@@ -21,31 +22,54 @@ public ISystemOperationsTests()
         public void DockerService_IsRunning()
         {
             var services = ServiceController.GetServices();
-            using (var dockerService = services.SingleOrDefault(service => service.ServiceName == "docker"))
+            using (var dockerService = services.FirstOrDefault(service => service.ServiceName == "docker" || service.ServiceName == "com.docker.service"))
```
galvesribeiro, line 25: `ServiceController` is not cross platform. We should find another way to deal with this. Perhaps have a common fixture that uses `Process` to check if Docker is running or not. This will definitively throw `PlatformNotSupportedException` on non-Windows machines and is also is deprecated on .Net 5 (which I intend to migrate soon).

galvesribeiro (changes requested):
Just a few tiny changes and we are good.

Thanks for the contribution! I really appreciate that!

src/Docker.DotNet/Models/ContainerExecInspectResponse.Generated.cs
dgvives: It's output like that by specgen

https://github.com/dotnet/Docker.DotNet/blob/a5ee0611f37711ec81fbdb5dc48f83e642da2a2a/tools/specgen/specgen.go#L601#L607

galvesribeiro: Ahhh yeah. I forgot it still Go-based. Sorry for the confusion, my branch doesn't use it anymore. Ignore what I said :) thanks!

test/Docker.DotNet.Tests/ISystemOperations.Tests.cs
dgvives: @galvesribeiro this test isn't executed in Windows platform because of the [SupportedOSPlatformsFact(Platform.Windows)] attribute. 
I agree, the propper thing would be for the test case to work on all supported platforms, but right now I have no idea on how to do it.
Do you have any suggestion? I'd appreciate that

galvesribeiro: Sorry, I've only looked to the snipped that changed.

You can replace the `ServiceController.GetServices()` with something like `Process.GetProcesses().FirstOrDefault(service => ....)`.

test/Docker.DotNet.Tests/ISystemOperations.Tests.cs
```diff
@@ -18,15 +18,11 @@ public ISystemOperationsTests()
...
+        [Fact]
+        public void Docker_IsRunning()
         {
-            var services = ServiceController.GetServices();
-            using (var dockerService = services.FirstOrDefault(service => service.ServiceName == "docker" || service.ServiceName == "com.docker.service"))
-            {
-                Assert.NotNull(dockerService); // docker is not running
-                Assert.Equal(ServiceControllerStatus.Running, dockerService.Status);
-            }
+            var dockerProcess = Process.GetProcesses().FirstOrDefault(_ => _.ProcessName == "docker" || _.ProcessName == "com.docker.service");
```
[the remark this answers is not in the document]
galvesribeiro: It is `docker` on OSX/Linux and `docker.exe` on Windows.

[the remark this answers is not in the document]
dgvives: Updated test case.

src/Docker.DotNet.X509/Docker.DotNet.X509.csproj
dgvives: After removing these, tests are failing in CI/CD pipeline and also 'dotnet build' from command line

src/Docker.DotNet/Endpoints/StreamUtil.cs
dgvives: Should be already updated

src/Docker.DotNet/JsonBase64Converter.cs
dgvives: Done

test/Docker.DotNet.Tests/ISystemOperations.Tests.cs
dgvives: Done

dgvives:
@galvesribeiro   would you have a look at the CI/CD pipeline and build operations? It might be slightly different to include Directory.Build.props, I can't build or test using 'dotnet build' or 'dotnet test' from command here

test/Docker.DotNet.Tests/ISystemOperations.Tests.cs
```diff
@@ -17,35 +18,54 @@ public ISystemOperationsTests()
...
+        [Fact]
+        public void Docker_IsRunning()
         {
-            var services = ServiceController.GetServices();
-            using (var dockerService = services.SingleOrDefault(service => service.ServiceName == "docker"))
-            {
-                Assert.NotNull(dockerService); // docker is not running
-                Assert.Equal(ServiceControllerStatus.Running, dockerService.Status);
-            }
+            var dockerProcess = Process.GetProcesses().FirstOrDefault(_ => _.ProcessName == "docker");
```
galvesribeiro, line 24: Now I'm not sure it will work on Windows as well 😄 

It should check for `docker` and `docker.exe` in order to support all OSes. Or just use `Contains()` and `IgnoreCase`.

galvesribeiro (changes requested):
Just one last thing.

galvesribeiro:
> Looks like updating Nerbank.GitVersioning package did the trick to compile using 'dotnet build'

Yeah, was about to say that. Github Actions changed the way the variables are added so they had to update the action to support that. Thanks!

test/Docker.DotNet.Tests/ISystemOperations.Tests.cs
galvesribeiro: Yeah, `Contains()` would be better.

galvesribeiro: For sure `Contains()`. I just confirmed that on _real_ docker services the process is called `dockerd` as @jterry75  mentioned. 

However, on desktops it is called `docker` or `docker.exe` depending on the OS. So `Contains()` would work just fine for all the cases.

galvesribeiro (approved):
LGTM. Thank you very much for the contribution and your patience!

galvesribeiro:
@jterry75 Its all good to me. Any final thoughts? Otherwise I'll squash and merge to start the release of a new package.

dgvives:
Next step: I'd like to order and cleanup the code base, applying Roslyn suggestions and removing warnings, to have it tidy and clean before continuing adding pending contributions/functionality. 
No modifications in classes, methods, names, namespaces, none of that whatsoever.

src/Docker.DotNet/Endpoints/StreamUtil.cs
dgvives: @jterry75 It looks like there was a reason behind disposing the stream.

Issue #236, opened 2017-07-22
CancellationTokenSource not work


303248153:
I need to stop the container after it runs longer than the time I give, then I try to use CancellationTokenSource, here is my code:

``` csharp
public static async Task DockerCancell()
{
	var cert = new CertificateCredentials(new X509Certificate2(CertPath, CertPassword));
	var configuration = new DockerClientConfiguration(new Uri("https://docker-1:2376"), cert);
	var client = configuration.CreateClient();
	var createResult = await client.Containers.CreateContainerAsync(new CreateContainerParameters()
	{
		Image = Image,
		Name = Guid.NewGuid().ToString(),
		WorkingDir = "/workdir",
		HostConfig = new HostConfig() { AutoRemove = true },
		Cmd = new[] { "sleep", "3" }
	});
	await client.Containers.StartContainerAsync(createResult.ID, new ContainerStartParameters());
	var token = new CancellationTokenSource();
	token.CancelAfter(1000);
	await client.Containers.WaitContainerAsync(createResult.ID, token.Token);
	Console.WriteLine("shouldn't be here");
}
```

It always output "shouldn't be here", looks like cancellation token didn't work for this api.
Further more, not only `CancelAfter`, `Cancel` is not work too:

``` csharp
public static async Task DockerCancell()
{
	var cert = new CertificateCredentials(new X509Certificate2(CertPath, CertPassword));
	var configuration = new DockerClientConfiguration(new Uri("https://docker-1:2376"), cert);
	var client = configuration.CreateClient();
	var createResult = await client.Containers.CreateContainerAsync(new CreateContainerParameters()
	{
		Image = Image,
		Name = Guid.NewGuid().ToString(),
		WorkingDir = "/workdir",
		HostConfig = new HostConfig() { AutoRemove = true },
		Cmd = new[] { "sleep", "3" }
	});
	await client.Containers.StartContainerAsync(createResult.ID, new ContainerStartParameters());
	var token = new CancellationTokenSource();
	// token.CancelAfter(1000);
	var task = client.Containers.WaitContainerAsync(createResult.ID, token.Token);
	await Task.Delay(1);
	token.Cancel();
	await task;
	Console.WriteLine("shouldn't be here");
}
```

jterry75:
I'm not sure what behavior you want here. When you cancel the task is allowed to have a continuation thus printing your `Console.WriteLine`. Are you expecting the library to throw in this case?

303248153:
[...]

PR #485, opened 2021-01-30
Feature/code cleanup

dgvives:
- Ordered and formatted code to ease adding pending PRs functionality.
- Removed most of the warnings when possible.
- Annotated some ArgumentNullException(object.property) to qualify full parameter name

My goal is to continue incorporating functionality over this base, gathering from existing PRs.
This PR does not add/modify/remove any existing functionality.


jterry75:
Hey @dgvives. Again super appreciate your work here. Do you think you could do a few systematic PR's with a single concept at once and then keep rebasing this one until we make sure all the changes were taken. So for example:
1 - Remove instances of this.
2 - Convert CancellationToken = default
3 - Fix spaces in files
4 - etc.

Make it so much easier to review. We can be timely on the reviews since they are very systematic and less error prone. What do you think?

dgvives:
@jterry75 totally agree, doesn't apply if there is no improvement. 

I'm working on #236 solution, just need a docker image valid to run tests on different environments win/unix/mac

PR #487, opened 2021-02-03
CancellationToken not cancelling

### LINKED ISSUE #236: CancellationTokenSource not work

### LINKED ISSUE #375: Response stream corruption when a request is cancelled

### LINKED ISSUE #416: Docker client 'defaultTimeout' not working
Using the following code, I would expect an exception to be raised in 5 seconds. Instead I am getting a OperationCanceledException at 20 seconds.

If I remove the defaultTimeout argument from the DockerClientConfiguration constructor, I no longer get an OperationCanceledException, but I do get a HttpRequestException after 20 seconds.

No matter what I change, an exception is always raised at 20 seconds. This is the code I am using (the IP address will be set to a host that is offline/unavailable).

Using Docker.DotNet 3.125.2 / .Net Core 2.2.

```
var client = new DockerClientConfiguration( new Uri( $"http://X.X.X.X:2375" ), defaultTimeout: TimeSpan.FromSeconds( 5 ) ).CreateClient();

var containers = await client.Containers.ListContainersAsync( new ContainersListParameters
{
	All = true
},
default );
```

### LINKED ISSUE #434: Events monitoring stream does not seems to always include CR or LF character at the end of every messages

dgvives:
Fixes #236 
Fixes #375
Fixes #416
Fixes #434
Closes #343 
Closes #351 
Closes #394

It solves execution stop on `DockerClient.Containers.WaitContainerAsync` waiting for the task to finish without being able to cancel.

```csharp
var waitContainerTask = dockerClient.Containers.WaitContainerAsync(createContainerResponse.ID, cts.Token);
```

- Task can be cancelled using token or default timeout.
- Included tests to verify cancellation works both ways.
- Included tests to several endpoints calling this method.
- Fixed tests failing on different execution contexts due to new exceptions being thrown.


dgvives:
Working solution pending tests cases.

dgvives:
@galvesribeiro @jterry75 and other maintainers, 

I'm developing integration tests for [IContainerOperations](https://github.com/dotnet/Docker.DotNet/blob/8e3cf4f7ac386e2f76791afc2c427a7b7884cd5e/src/Docker.DotNet/Endpoints/IContainerOperations.cs#L10) endpoint, since I need to test this piece of code via `IContainerOperations` endpoint. 

It would be better to just develop tests for affected functionality, right?
And include tests for currently untested endpoint methods on a different changeset, to keep it as small as possibletests

dgvives:
Closed, tests produce different results on each execution

dgvives:
Updated tests to include timeout

dgvives:
Included tests to validate issues solution.

dgvives:
Core changes are set to a minimum. However, some issues are complicated to test.
New test cases have been added to validate issue solution, and a common test fixture put in place to avoid collisions between test monitoring events.
All core modifications have been tested and tests are included within this PR.

@jterry75 @galvesribeiro would you have a look on it?

src/Docker.DotNet/Endpoints/StreamUtil.cs
```diff
@@ -10,37 +10,44 @@ namespace Docker.DotNet.Models
...
-                // ReadLineAsync must be cancelled by closing the whole stream.
-                using (cancel.Register(() => stream.Dispose()))
+                using (var reader = new StreamReader(stream, new UTF8Encoding(false)))
                 {
-                    using (var reader = new StreamReader(stream, new UTF8Encoding(false)))
+                    while (!cancel.IsCancellationRequested)
                     {
-                        string line;
-                        while ((line = await reader.ReadLineAsync()) != null)
+                        string line = await await Task.WhenAny(reader.ReadLineAsync(), tcs.Task);
```
jterry75, line 26: Fascinating idea. You seem to have an extra `await` though.

src/Docker.DotNet/Endpoints/StreamUtil.cs
```diff
@@ -10,37 +10,44 @@ namespace Docker.DotNet.Models
 {
     internal static class StreamUtil
     {
-        private static Newtonsoft.Json.JsonSerializer _serializer = new Newtonsoft.Json.JsonSerializer();
+        private static readonly Newtonsoft.Json.JsonSerializer _serializer = new Newtonsoft.Json.JsonSerializer();
 
         internal static async Task MonitorStreamAsync(Task<Stream> streamTask, DockerClient client, CancellationToken cancel, IProgress<string> progress)
         {
+            var tcs = new TaskCompletionSource<string>();
+            cancel.Register(() => tcs.TrySetCanceled(default));
```
jterry75, line 18: Why not in a using like the others?

jterry75 (commented):
Just a couple of questions. Looks good

src/Docker.DotNet/Endpoints/StreamUtil.cs
dgvives: Should be like others by next commit. I'm running tests now before pushing again

src/Docker.DotNet/Endpoints/StreamUtil.cs
dgvives: Yes, it looks like that.

`await Task.WhenAny(task1, task2)` would return either task1 or task2.
Then, this task1 or task2 need to be awaited too.

Original idea from #343 by @suneption, I just adjusted it for tests to pass.

jterry75: Crazy. I did not know you could inline await :).

src/Docker.DotNet/Endpoints/StreamUtil.cs
```diff
@@ -1,97 +1,61 @@
...
+        {
+            var tcs = new TaskCompletionSource<string>();
+
+            using (var stream = await streamTask)
+            using (var reader = new StreamReader(stream, new UTF8Encoding(false)))
+            using (cancel.Register(() => tcs.TrySetCanceled(default)))
+            {
+                while (!cancel.IsCancellationRequested)
+                {
+                    string line = await await Task.WhenAny(reader.ReadLineAsync(), tcs.Task);
```
jterry75, line 25: Alright one last question. When `tcs.Task` is selected the 2nd `await` will throw because there is no result. I assume this is the correct behavior and we want to throw the cancellation exception?

jterry75 (approved):
LGTM

jterry75:
@galvesribeiro - Will you check my last question about throwing the cancellation exception here. As long as your good with that we can take this change IMO.

src/Docker.DotNet/Endpoints/StreamUtil.cs
dgvives: When cancelling a token I expect to get an OpertionCancelledException. Otherwise it would mean task reached its end instead of being cancelled. 

I have no knowledge about architecture or system designs, i just enjoy thinks working on the way I like. Do the minimum to not do it twice.

I've included several tests on this, expecting OperationCancelledException to be thrown and failing test otherwise. 
There should be on IContainerOperationsTests

galvesribeiro: I don't think it will throw there. The `tcs.Task` will be returned.  

In this case I wouldn't assign to the `line` variable directly. I would assign it to a Task and would check which task is that - the readline or the cancellation. 

If that is the cancellation task, then yeah, it is cancelled. If it is the deadline Task, then you `string line = await readLineTask();`.

dgvives: There should be a test behind this

galvesribeiro: Agreed.

dgvives:
@jterry75 @galvesribeiro  to not merge this. Tested behavior is not right on `MonitorStreamAsync`

`MonitorStreamForMessagesAsync`, `MonitorResponseForMessagesAsync`, and `MonitorStreamAsync` behave in the same way, throwing **TaskCancelledException** when task is cancelled. 

Expect unified behavior to be pushed soon

dgvives:
image

Some tests have output details, I used them to understand what I was doing. At some point I gave up but test output details remain

test/Docker.DotNet.Tests/IContainerOperationsTests.cs
```diff
@@ -0,0 +1,558 @@
...
+                _cts.Token
+                );
+
+            _output.WriteLine($"Line count: {logList.Count}");
+
+            Assert.NotEmpty(logList);
+        }
+
+        //  @dgvives: This test doesn't work on xUnit test project. Execution never finishes.
+        //            Same code on consoleApp works just fine
```
Emdot, line 131: This may suggest a deadlock bug in the code, rather than an xUnit problem. It's common for async code to work in Console apps but break elsewhere, due to the difference in synchronization contexts. Unfortunately, I'm not an expert in how sync contexts and schedulers and such work, so I don't know where to suggest looking.

src/Docker.DotNet/DockerClient.cs
```diff
@@ -346,22 +346,22 @@ internal DockerClient(DockerClientConfiguration configuration, Version requested
...
-            // If there is a timeout, we turn it into a cancellation token. At the same time, we need to link to the caller's
-            // cancellation token. To avoid leaking objects, we must then also dispose of the CancellationTokenSource. To keep
-            // code flow simple, we treat it as re-entering the same method with a different CancellationToken and no timeout.
+            var request = PrepareRequest(method, path, queryString, headers, data);
+
             if (timeout != s_InfiniteTimeout)
             {
                 using (var timeoutTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken))
                 {
                     timeoutTokenSource.CancelAfter(timeout);
```
Emdot, line 355: Would it make sense to move the timeout complexity to a separate function? Then this whole method can be:
```
var request = PrepareRequest(method, path, queryString, headers, data);
return RunWithTimeoutAsync(
    ct => _client.SendAsync(request, completionOption, ct), 
    cancellationToken, timeout);
```

A RunWithTimeoutAsync implementation might look like this:
```
public static async Task<T> TimeoutAfter<T>(Func<CancellationToken, Task<T>> makeTask, CancellationToken inputToken, TimeSpan timeout)
{
    var cts = new CancellationTokenSource();

    var timerTask = Task.Delay(timeout, cts.Token);
    var mainTask = makeTask(cts.Token);
    
    // Start both the timeout and the main task in parallel.
    Task firstTask == await Task.WhenAny(timerTask, mainTask).ConfigureAwait(false);
    
    // At this point, one of them has finished. Cancel the other one.
    cts.Cancel();
    
    // If the timeout occurred first, throw a timed-out exception.
    if (firstTask == timerTask)
        throw new TimeoutException();
    
    // Otherwise return the result of the main task. If it was canceled, this may throw 
    // a cancelation exception. If it faulted, it may throw some other exception.
    return await mainTask.ConfigureAwait(false);
}
```

For added performance you could use a state parameter to avoid the closure. As a bonus, timeouts emit a separate exception than cancellation.

dgvives: @Emdot great suggestion. 
I'll wait for this to get approved and continue with improvements.

test/Docker.DotNet.Tests/IContainerOperationsTests.cs
dgvives: I see, I'm creating some sort of web app to test this. Will update with test results

dgvives: My bad, failing test wasn't **IContainerOperations.GetContainerLogs_Tty_True_StreamReadsLogs** but **IContainerOpreations.GetContainerStatsAsync_StreamStats**

I didn't notice it because it works fine on build agent.

dgvives: JsonReader.ReadAsync can't be cancelled once serialization has started.
Looking for an alternative now

dgvives:
@jterry75 moving back to stream.Dispose() solved cancellation problems while monitoring streams. 
Tests have been added to validate failing tasks under different circumstances

dgvives:
Closed while research on stream.read not being cancelled is done

src/Docker.DotNet/Endpoints/StreamUtil.cs
dgvives: I listened to your advice. It's not exactly as you said but shares the same idea.
It helped fixing some problems when reading streams from containers created without a tty
Thank you @galvesribeiro

dgvives:
When using `JsonSerializer.Deserialize<T>(JsonReader)` to deserialize, deep inside it is calling `ExpandoObject.ReadReadObject(JsonReader)` and no cancellation is possible from inside the loop, there is no token to cancel. 


I went over the source for JsonReader, StreamReader, Stream, and finally JsonSerializer until I found out this.


https://github.com/JamesNK/Newtonsoft.Json/blob/b6dc05be5a0f4808f06ec430f3bb59b24d3fbc3e/Src/Newtonsoft.Json/Converters/ExpandoObjectConverter.cs#L113#L141
```csharp
        private object ReadObject(JsonReader reader)
        {
            IDictionary<string, object?> expandoObject = new ExpandoObject();

            while (reader.Read())
            {
                switch (reader.TokenType)
                {
                    case JsonToken.PropertyName:
                        string propertyName = reader.Value!.ToString();

                        if (!reader.Read())
                        {
                            throw JsonSerializationException.Create(reader, "Unexpected end when reading ExpandoObject.");
                        }

                        object? v = ReadValue(reader);

                        expandoObject[propertyName] = v;
                        break;
                    case JsonToken.Comment:
                        break;
                    case JsonToken.EndObject:
                        return expandoObject;
                }
            }

            throw JsonSerializationException.Create(reader, "Unexpected end when reading ExpandoObject.");
        }
```



This hasn't been to easy to find out, difficult to figure out a solution, and turned even more complicated to test.

dgvives:
@jterry75 apologies as I've been up and down with this PR. It turned out to be way more complicated than it seemed
@galvesribeiro thanks for your suggestion, please take a look at how it ended up

I'm having some rest after this

jterry75:
@dgvives - Looks like you got it!

dgvives:
@jterry75 - Yes, it has been a very interesting journey into task cancelling and tests

jterry75:
Do you want to rebase all your commits into one just because it was a lot of change over them and then I'll do a final review and signoff. But looks good to me

dgvives:
Done

dgvives:
@jterry75 done

jterry75 (approved):
LGTM

jterry75:
@galvesribeiro - Want to take a look as well

galvesribeiro (approved):
Awesome work! Thank you for the contribution and the huge patience across the reviews.

galvesribeiro:
@jterry75 do you think we have enough for a new release now?

galvesribeiro:
@jterry75 ?

PR #494, opened 2021-02-17
CreateContainerAsync: Throw DockerImageNotFoundException when image not found

Emdot:
`ContainerOperations.CreateContainerAsync` now throws `DockerImageNotFoundException` instead of `DockerContainerNotFoundException `when the image can't be found.

### DIFF
--- a/test/Docker.DotNet.Tests/IContainerOperationsTests.cs
+++ b/test/Docker.DotNet.Tests/IContainerOperationsTests.cs
@@ -1,0 +1,34 @@
+﻿using System;
+using System.Threading.Tasks;
+using Docker.DotNet.Models;
+using Xunit;
+
+namespace Docker.DotNet.Tests
+{
+	public class IContainerOperationsTests : IDisposable
+	{
+		private readonly DockerClient _client;
+
+		public IContainerOperationsTests()
+		{
+			_client = new DockerClientConfiguration().CreateClient();
+		}
+
+		[Fact]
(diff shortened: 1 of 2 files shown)


test/Docker.DotNet.Tests/IContainerOperationsTests.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
```
dgvives, line 1: @Emdot nice job. 
Previous exception thrown was DockerContainerNotFound I suppose, could be more errors based on response HttpStatusCode or similar?
I might check this for some tests I did myself.

👍

jterry75:
This was a regression right?

test/Docker.DotNet.Tests/IContainerOperationsTests.cs
Emdot: The API docs for the endpoint are wrong too, so presumably that's the source of the mistake.

Emdot:
Not that I'm aware of, but I didn't dig into it. What makes you think that?

jterry75:
I'm surprised that the Docker API has this bug and is unreported. I think I agree that 404 is incorrectly documented here and you have the right code. At present this code is doing "what the api documents" as you suggest. So I'm just wondering if you know that Docker did this wrong or did it intentionally?

Emdot:
Let's find out! https://github.com/moby/moby/pull/42047

PR #496, opened 2021-02-18
ISystemOperations.AuthenticateAsync now returns AuthResponse.

Emdot:


jterry75 (approved):
LGTM

jterry75:
@galvesribeiro - PTAL

galvesribeiro (approved):
LGTM. Thanks!

galvesribeiro:
Just one thing before we merge. Why is the tests failing? I'm out of my desk now and I can't check. Can you please fix it and push the update? Thanks!

Emdot:
It's failing MonitorEventsAsync_Succeeds. Running locally, it appears to be an intermittent failure.

dgvives:
@galvesribeiro @Emdot  errors on **StreamUtils.MonitorStreamAsync** methods are not related to this changeset but to previously existing changes.

Previous [ContainerOperations.GetContainerLogsAsync](https://github.com/dotnet/Docker.DotNet/blob/0d13c419df2ecd64570bb4c3ce73e978c5e91bf6/src/Docker.DotNet/Endpoints/StreamUtil.cs#L20) task cancellation by disposing stream had some pitfalls, as it was disposing a stream while a streamReader was consuming from it.

Should be working fine now, pushed to #487  .
@Emdot feel free to review/fix/improve/criticise this PR. There is integrity within your criteria

Emdot:
@galvesribeiro, can you retrigger the build?

galvesribeiro:
Just did. Same error as it seems... ☹️

Emdot:
I see that this is a known bug in the test case. "[// On CI/CD Pipeline exception is thrown, not always](https://github.com/dotnet/Docker.DotNet/blob/master/test/Docker.DotNet.Tests/ISystemOperations.Tests.cs#L212)"

I think there's more of a problem here than I'm willing to fix:
1. It _looks_ like the event streams are delayed. Some of the events from CreateImageAsync don't arrive until after MonitorEventsAsync starts, so they [show up in _onJSONMessageCalled](https://github.com/dotnet/Docker.DotNet/pull/496/checks#step:5:16). Some of the events from TagImageAsync don't arrive until [after the assertion](https://github.com/dotnet/Docker.DotNet/pull/496/checks#step:5:39).
2. I can't be _sure_ that that's what's happening, since stdio might be buffered too. Adding some Console.Out.Flush and Console.Error.Flush calls between commands can help check for that.
3. This _probably_ isn't what's going wrong in this case, but it's a bug anyway: Using MonitorEventsAsync this way has a race condition. Unless you `await` MonitorEventsAsync's Task, you can't guarantee that the monitoring has started, so you can miss messages. The fix would is to replace MonitorEventsAsync with something like StartEventMonitorAsync that returns a subscription.

Would you accept a PR to disable the broken test case, until someone fixes the underlying problems?

Emdot:
I wrote that up as #502.

dgvives:
@Emdot move the call to include a call to remove the image after tagging it, that should be enough in this case

```csharp
await _client.Images.CreateImageAsync(new ImagesCreateParameters { FromImage = "hello-world" }, null, progressJSONMessage);
var task = Task.Run(() => _client.System.MonitorEventsAsync(new ContainerEventsParameters(), progressMessage, cts.Token));
await _client.Images.TagImageAsync(repository, new ImageTagParameters { RepositoryName = repository, Tag = newTag });
await _client.Images.DeleteImageAsync($"{repository}:{newTag}", new ImageDeleteParameters());
cts.Cancel();

```

Emdot:
I'm unwilling to fudge the tests to make them pass. If you fix the problematic code (as noted above) or disable the test, then this PR can proceed.

PR #500, opened 2021-02-26
Generate sdk docs

Emdot:
Sets up actions to generate HTML documentation for the SDK project. You can see what it looks like [here](https://emdot.github.io/Docker.DotNet/).

Note: Presently the docs are published at every pull request. When it's time to debut the SDK layer we'll want to move that task from ci.yml to publish.yml.


Emdot:
Something's wrong--this is pulling too much. Closing the request and will redo it.

PR #503, opened 2021-03-05
Fix testFixture

dgvives:
Bug in swarm initialization for tests


dgvives:
It won't restore previous swarm state if it was uninitialized

jterry75 (approved):
LGTM

Issue #502, opened 2021-02-27
Docker.DotNet.Tests.ISystemOperationsTests.MonitorEventsAsync_Succeeds test fails


Emdot:
**Steps to reproduce the issue:**
1. Submit a PR
2. Sometimes the CI build fails

**What actually happened?:**
Test case fails despite no changes to relevant code. See https://github.com/dotnet/Docker.DotNet/pull/496/checks?check_run_id=1992119140 for an example.

**What did you expect to happen?:**
Test case passes.

**Additional information:**
I see that this is a known bug in the test case. "[// On CI/CD Pipeline exception is thrown, not always](https://github.com/dotnet/Docker.DotNet/blob/master/test/Docker.DotNet.Tests/ISystemOperations.Tests.cs#L212)"

I think there's more of a problem here than I'm willing to fix:
1. It _looks_ like the event streams are delayed. Some of the events from CreateImageAsync don't arrive until after MonitorEventsAsync starts, so they [show up in _onJSONMessageCalled](https://github.com/dotnet/Docker.DotNet/pull/496/checks#step:5:16). Some of the events from TagImageAsync don't arrive until [after the assertion](https://github.com/dotnet/Docker.DotNet/pull/496/checks#step:5:39).
2. I can't be _sure_ that that's what's happening, since stdio might be buffered too. Adding some Console.Out.Flush and Console.Error.Flush calls between commands can help check for that.
3. This _probably_ isn't what's going wrong in this case, but it's a bug anyway: Using MonitorEventsAsync this way has a race condition. Unless you `await` MonitorEventsAsync's Task, you can't guarantee that the monitoring has started, so you can miss messages. The fix would is to replace MonitorEventsAsync with something like StartEventMonitorAsync that returns a subscription.

PR #549, opened 2022-01-27
Dispose the response object in case of errors

nertim:
In the existing implementation on an error response an exception is thrown. Due to this the original response object is not disposed leaving connections open. This change specifically disposes the response object before an exception is thrown either by one of the handlers OR the generic dockerApiException.

NOTE: The callers could not be wrapped around a `using` statement due to the return of an async stream task.

### DIFF
--- a/src/Docker.DotNet/DockerClient.cs
+++ b/src/Docker.DotNet/DockerClient.cs
@@ -270,7 +270,7 @@ namespace Docker.DotNet
 
             await HandleAndDisposeIfErrorResponseAsync(response.StatusCode, response, errorHandlers);
 
-            return await response.Content.ReadAsStreamAsync().ConfigureAwait(true);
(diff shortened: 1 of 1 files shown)


EricSten-MSFT (approved):

nertim:
Hello @jterry75, @galvesribeiro,

I am unable to repro this test failure locally on my box. I also wonder if this is somehow related to the existing issue #502 

Could you please provide some guidance on the best way for me to repro this locally and/or to proceed?

Thank you :)

galvesribeiro:
Is this still a thing? If yes, can we rebase on main so we can merge it?

Thanks!

galvesribeiro:
Ok, I haven't got any updates on this and I can't reproduce issue, so I'm going to close it. Feel free to reopen and rebase if you can make a repro with a test.

Thanks!

Issue #554, opened 2022-01-29
cannot hijack chunked or content length stream


Alex-1357:
Hello, why this module https://github.com/dotnet/Docker.DotNet/blob/master/src/Docker.DotNet/Microsoft.Net.Http.Client/HttpConnectionResponseContent.cs said me - "cannot hijack chunked or content length stream".

This is my attempt to attach to container

    Async Sub AttachToContainer(DockerHub As DockerClient, ContainerID As String, Cmd As IList(Of String), ExecCts As CancellationToken)
        Dim ExecTask = Await DockerHub.Exec.ExecCreateContainerAsync(ContainerID, New ContainerExecCreateParameters With {
                         .AttachStdin = False,
                         .AttachStderr = True,
                         .AttachStdout = True,
                         .Tty = False,
                         .Cmd = Cmd
                       },
                       ExecCts)
        Try
            Dim Stream = Await DockerHub.Exec.StartAndAttachContainerExecAsync(ExecTask.ID, False)
            Await Stream.CopyOutputToAsync(
                    Nothing,
                    Console.OpenStandardOutput(),
                    Console.OpenStandardError(),
                    CancellationToken.None)
        Catch ex As Exception
            Console.WriteLine(ex.Message)
        End Try
    End Sub

I'm don't sure that this approach is right, I need to realize simple operation

    # sudo docker exec -it 7dde487b4424 pwd

But receive this result

![docker_01302022_015254-1](https://user-images.githubusercontent.com/86534217/151681485-307a74b1-acf8-44e6-a413-22d717ebf291.png)

wesley-pattison:
Hi, I'd like to also mention I am receiving this exception. My case is a bit different:

- My code is in net 6.0 C#
- I'm using [Rancher Desktop](https://github.com/rancher-sandbox/rancher-desktop) with the dockerd (moby) option checked for Container Runtime, WSL Integration is also enabled.
- trying to leverage [dotnet-testcontainers](https://github.com/HofmeisterAn/dotnet-testcontainers) nuget package which underneath the hood, uses the Docker DotNet library.

I asked a colleague of mine who still had Docker Desktop installed, to try the example unit test they dotnet-testcontainers project supplies, and it works. Unfortunately using Docker Desktop for Windows is not an option due to licensing.

Some debugging information that may help:

This is the code in question that throws the exception.
```
        public WriteClosableStream HijackStream()
        {
            if (_responseStream != _connection.Transport)
            {
[...]

PR #566, opened 2022-06-24
#554 Cannot hijack chunked or content length stream

joebone:
…h Rancher Desktop ( http 1.0 doesn't allow chunked encoding )

### DIFF
--- a/src/Docker.DotNet/DockerClient.cs
+++ b/src/Docker.DotNet/DockerClient.cs
@@ -7,6 +7,7 @@ using System.Net;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
+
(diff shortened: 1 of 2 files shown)


HofmeisterAn:
I think we should not fall back to HTTP 1.0. Instead, I suggest fixing the `ChunkedReadStream` implementation (https://github.com/dotnet/Docker.DotNet/issues/554#issuecomment-1170444866), respectively support chunked transfer encoding. If you look into the Docker API documentation, you'll see the explicit return of HTTP 1.1.

joebone:
Sure, appreciate the feedback. For the time being we'll keep using it as a workaround, but look forward to the proper chunked implementation when it arrives 👍

PR #567, opened 2022-06-29
Strong name the assemblies

### LINKED ISSUE #403: Can you provide a strong name for the Nuget package?
Could not load file or assembly 'Docker.DotNet, Version=3.125.2.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. A strongly-named assembly is required. (Exception from HRESULT: 0x80131044)

JamieMagee:
Key generated using `sn -k Key.snk`

Closes #403


dfederm (approved):

JamieMagee:
@galvesribeiro can you review this please?

JamieMagee:
@galvesribeiro is it possible to get this included in the next release please?

HofmeisterAn (approved):
@galvesribeiro I think we can merge this. Not having the strong name prevents .NET Framework projects using the library in some cases.

HofmeisterAn:
@galvesribeiro Can we merge this, please?

galvesribeiro:
Ok. Got it. I thought SN wasn't a requirement on current .Net framework versions and only old ones used it.

It is important to make clear that this library, after the refactory in the works, will only support the supported versions of .Net. Anything out of support policy will not be included. 

For now, lets merge this.

Thanks!

HofmeisterAn:
> It is important to make clear that this library, after the refactory in the works, will only support the supported versions of .Net. Anything out of support policy will not be included.

Why is that? Are there any requirements or reasons to retarget from .NET Standard to .NET 5+? Do we really need the runtime features and drop support for .NET Framework? The .NET Framework 4.6.2 EOL is 2027 (4.8.1 TBD) wich in theory supports .NET Standard 2.0.

In general. I am fine with .NET 5+, but we exclude developers from using Docker.DotNet then (something I would like to avoid).

galvesribeiro:
Yeah sorry. That is what I meant. If that is a supported version so be it.

In the other hand, when/if the TFMs are really simplified for net8 ownward, that will remove the old frameworks.

Anyway, nothing changes now. It is just a heads up that we will not hold back because of old .net framework.

Sent from my iPhone

On Dec 4, 2022, at 6:24 AM, Andre Hofmeister ***@***.***> wrote:

﻿

It is important to make clear that this library, after the refactory in the works, will only support the supported versions of .Net. Anything out of support policy will not be included.

Why is that? Are there any requirements or reasons to retarget from .NET Standard to .NET 5+? Do we really need the runtime features and drop support for .NET Framework? The .NET Framework 4.6.2 EOL is 2027 (4.8.1 TBD) wich in theory supports .NET Standard 2.0.

In general. I am fine with .NET 5+, but we exclude developers from using Docker.DotNet then (something I would like to avoid).

HofmeisterAn:
Blocks the OpenTelemetry in .NET folks from using it...

JamieMagee:
@HofmeisterAn just to clarify, the lack of this change being released is blocking you?

HofmeisterAn:
> just to clarify, the lack of this change being released is blocking you?

Yep, breaks the OTel build:

    CSC : warning CS8002: Referenced assembly 'Docker.DotNet, Version=3.125.0.0, Culture=neutral, PublicKeyToken=null' does not have a strong name.

galvesribeiro:
Sorry for all the delay. Spoke on Slack with @HofmeisterAn. The new package was pushed which include the strong naming as per this PR.

Will discuss next steps from there. 

Thanks!

PR #577, opened 2022-07-22
Update ContainerLogsParameters.Generated.cs

13750573877:
Added missing until parameter


HofmeisterAn:
I assume this should be done via `specgen` for a specific API version. The model classes are generated.

galvesribeiro:
Yes, you need to update the specimen and not those files.

HofmeisterAn:
I'll prepare the update with `specgen`. @galvesribeiro I noticed that e.g. this model class is part of `modeldefs.go` and not generated from the Docker API sources. Do I need to update those types by hand? Then I'll compare all types in the file agains the Docker API.

galvesribeiro:
Yep that is the idea.

PR #586, opened 2022-09-06
Fix #518, #522

### LINKED ISSUE #518: HttpClient no dispose
https://github.com/dotnet/Docker.DotNet/blob/e51de97addebec71c1b1dd7f399f06e381a1044c/src/Docker.DotNet/DockerClient.cs#L450

### LINKED ISSUE #522: Docker.DotNet.Models.ImageBuildParameters has missing and extra properties
Docker.DotNet version 3.125.0.0

Class Docker.DotNet.Models.ImageBuildParameters is supposed to match the request for Docker API's `/build` endpoint; however, it doesn't match the current version (v 1.41) or any recent version (I checked back to 1.25). The class is missing some properties and has several extra properties.

Missing:
* pull
* outputs

Extra:
* securityopt
* ulimits
* cgroupparent
* session
* cpusetmems
* isolation
* pullparent

HofmeisterAn:
@galvesribeiro The first commit disposes the `HttpClient`. Usually, you would like to keep a single instance of the `HttpClient` for the entire application. I am not certain if that works with the current implementation. That's why I kept it as a non static field, disposing it makes sense then.

The second one removes the extra properties. I double-checked them against the API docs. `pull` and `outputs` where part of e43cdd96ef089ec20f99defb5b83b047e2bdec04 and detected by my script (as I said, I didn't really check the other way around).

Closes #518, closes  #522.

### DIFF
--- a/src/Docker.DotNet/DockerClient.cs
+++ b/src/Docker.DotNet/DockerClient.cs
@@ -151,6 +151,7 @@ namespace Docker.DotNet
         public void Dispose()
         {
             Configuration.Dispose();
+            _client.Dispose();
         }
 
         internal Task<DockerApiResponse> MakeRequestAsync(
--- a/src/Docker.DotNet/Models/ImageBuildParameters.Generated.cs
+++ b/src/Docker.DotNet/Models/ImageBuildParameters.Generated.cs
@@ -24,20 +24,11 @@ namespace Docker.DotNet.Models
         [QueryStringParameter("forcerm", false, typeof(BoolQueryStringConverter))]
         public bool? ForceRemove { get; set; }
 
-        [QueryStringParameter("pullparent", false, typeof(BoolQueryStringConverter))]
-        public bool? PullParent { get; set; }
-
         [QueryStringParameter("pull", false)]
         public string Pull { get; set; }
 
-        [QueryStringParameter("isolation", false)]
-        public string Isolation { get; set; }
-
         [QueryStringParameter("cpusetcpus", false)]
         public string CPUSetCPUs { get; set; }
-
-        [QueryStringParameter("cpusetmems", false)]
-        public string CPUSetMems { get; set; }
 
         [QueryStringParameter("cpushares", false)]
         public long? CPUShares { get; set; }
@@ -54,9 +45,6 @@ namespace Docker.DotNet.Models
         [QueryStringParameter("memswap", false)]
         public long? MemorySwap { get; set; }
 
-        [QueryStringParameter("cgroupparent", false)]
-        public string CgroupParent { get; set; }
-
         [QueryStringParameter("networkmode", false)]
         public string NetworkMode { get; set; }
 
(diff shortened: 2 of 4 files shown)


HofmeisterAn:
The tests `CreateContainerAsync_TimeoutExpires_Fails` and `GetContainerLogs_Tty_False_Follow_False_ReadsLogs` seams flaky. I saw them failing on other pull requests too. I can take a look at them, try to find the root cause, and create another pull request with a fix in the next days.

HofmeisterAn:
I'll take a look into the broken tests.

galvesribeiro:
I wouldn't change the behavior of HttpClient on this release. This is very problematic given the way everything else works.

On the new/refactored code, We would use `HttpClientFactory` instead, and not manage the lifecycle of it anymore.

HofmeisterAn:
> I wouldn't change the behavior of HttpClient on this release. This is very problematic given the way everything else works.

Disposing the client will not change the behavior, it is just cleaner if someone disposes `DockerClient`. But if you like I can drop the commit. Regarding the tests:

1) `CreateContainerAsync_TimeoutExpires_Fails` does not make sense. We cannot control when `CreateContainerAsync` returns a successful result. The test is not deterministic.
2) `GetContainerLogs_*` are a bit messy. They share an identical setup. I think we can use a `Theory` in the future to test all kinds of combinations. Right now, we only test a few of them. I made certain that we await the `containerLogsTask` at the end of each test.

HofmeisterAn:
Are any adjustments necessary?

dgvives:
@HofmeisterAn is there any other way to check if the DockerClient timeout is being considered on each call? I couldn't come out with any alternative

HofmeisterAn:
@dgvives what do you mean by "any other way"? Looking into the `DockerClient` implementation, it looks like the (default) timeout is not consistent passed to the HTTP calls. Which probably makes sense (at least for calls that deal with streams).

dgvives:
@HofmeisterAn the default timeout is used to create a CancellationToken for the http request.

https://github.com/dotnet/Docker.DotNet/blob/master/src/Docker.DotNet/DockerClient.cs#L370-L378
```csharp
            if (timeout != SInfiniteTimeout)
            {
                using (var timeoutTokenSource = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken))
                {
                    timeoutTokenSource.CancelAfter(timeout);
                    return await _client.SendAsync(request, completionOption, timeoutTokenSource.Token)
                        .ConfigureAwait(false);
                }
            }
````

The test to verify that the request is cancelled after the specified timeout was done using a non-streamed call to DockerClient, in this case `CreateContainerAsync`. I wonder if there is another way to verify that the default timeout is taken into account in the requests.

HofmeisterAn:
Ah now I get it, you reffering to:

    There is nothing we can do to delay CreateContainerAsync (aka HttpClient.SendAsync) deterministic. We cannot control if it responses successful before the timeout.

Somewhere I had an example that behaves (partially) like an HTTP Docker endpoint (I used it to do some load tests). Maybe we can use this, to enable the test again. I will take a look into it (share it) in the next days. Have some nice holidays.

HofmeisterAn:
@dgvives I use something like [this](https://gist.github.com/HofmeisterAn/248b926ce2fc80d8118b54ba8788e840) to test or debug the Docker.DotNet implementation.

PR #590, opened 2022-09-21
Handle Reponse Error when building image

aDisplayName:
`MakeRequestForRawResponseAsync` is used when calling `IImageOperations.CreateImageAsync`. But the response error from daemon was not handled. Add the error handling before return the result.

Fix issue #555

PR #591, opened 2022-09-23
fix(#590): Do not pull non existing images in monitor event unit tests

HofmeisterAn:


HofmeisterAn:
@aDisplayName @galvesribeiro I think this is better. We do not need to pull images (that do not exist) to test the monitor event API.

galvesribeiro (approved):
LGTM

PR #599, opened 2022-11-28
Update ISystemOperations.cs

MagnusMat:
Changed Obselete warning to read out the correct shape of the MonitorEventsAsync Task.

PR #604, opened 2022-12-28
fix: Sync Docker system monitor events

HofmeisterAn:
## What does this PR do?

Synchronizes the Docker system event tests with the help of `AutoResetEvent`. At least two tests (`MonitorEventsAsync_IsCancelled_NoStreamCorruption` and `MonitorEventsFiltered_Succeeds`) do not synchronize the events. The tests are not deterministic, they might run into a race condition where events are not published on assertions yet.

## Why is it important?

Some `Docker.DotNet` tests are flaky, PR improves the reliability of the tests.

## Related issues

- Relates #599

## Follow-ups

- Looking closer into the tests, I noticed (gut feeling) events triggered, quickly after another might not be published reliable. It looks like they are swallowed.
- I think we can reduce the complexity of `StreamUtil`.
- /cc @dgvives You might want to take a look at the _stream corruption_ test changes.

### DIFF
--- a/src/Docker.DotNet/Endpoints/StreamUtil.cs
+++ b/src/Docker.DotNet/Endpoints/StreamUtil.cs
@@ -1,5 +1,4 @@
 using System;
-using System.Diagnostics;
 using System.IO;
 using System.Net.Http;
 using System.Text;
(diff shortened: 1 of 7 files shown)


dgvives:
@HofmeisterAn the _stream corruption_ tests come from #375, I didn't run into this problem myself but I was able to replicate it using that piece of code.
I would say tests are good if no stream corruption happens

HofmeisterAn:
> @HofmeisterAn the _stream corruption_ tests come from #375, I didn't run into this problem myself but I was able to replicate it using that piece of code. I would say tests are good if no stream corruption happens

I noticed that the events are not published reliable and fixed them too. If you e.g. look at the other PR's [build](https://github.com/dotnet/Docker.DotNet/actions/runs/3583328566/jobs/6052552952#step:5:139) (ff.) you see no event or message output. I got them sometimes on my local runs though.

HofmeisterAn:
@galvesribeiro can you please take a look into this PR? I would like to work on some follow-ups.

HofmeisterAn:
@galvesribeiro can you please review the PR?

PR #639, opened 2023-05-26
feat: Update to Docker v24.0.2

HofmeisterAn:
## What does this PR do?

The pull request updates Moby to its latest version, [v24.0.2](https://github.com/moby/moby/releases/tag/v24.0.2). A few classes have been renamed:

- https://github.com/moby/moby/commit/72938574568461ee5e544459a779c5fd4ac14103
- https://github.com/moby/moby/commit/3bb2d0026b9353e02a011fc64c821534bd1cae36
- https://github.com/moby/moby/commit/f19ef20a4414a6f63da8fb2493e6010359652ad1

This change https://github.com/moby/moby/commit/240a9fcb833996b1365266b906e181a1a6e06549 requires probably slightly more adjustments. `specgen` fails to generate structs because the file name clashes. Specifically the Info, Secret and Topology structs for swarm and volume.

## Related issues

\-

### DIFF
--- a/tools/specgen/specgen.go
+++ b/tools/specgen/specgen.go
@@ -37,7 +37,7 @@ var typesToDisambiguate = map[string]*CSModelType{
 			},
 		},
 	},
-	typeToKey(reflect.TypeOf(container.ContainerCreateCreatedBody{})): {Name: "CreateContainerResponse"},
+	typeToKey(reflect.TypeOf(container.CreateResponse{})): {Name: "CreateContainerResponse"},
 	typeToKey(reflect.TypeOf(container.HealthConfig{})): {
 		Properties: []CSProperty{
 			CSProperty{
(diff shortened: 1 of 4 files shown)


tools/specgen/specgen.go
```diff
@@ -70,20 +71,23 @@ var typesToDisambiguate = map[string]*CSModelType{
 			},
 		},
 	},
+	typeToKey(reflect.TypeOf(volume.Info{})):                 {Name: "VolumeInfo"},
+	typeToKey(reflect.TypeOf(volume.Topology{})):             {Name: "VolumeTopology"},
+	typeToKey(reflect.TypeOf(volume.Secret{})):               {Name: "VolumeSecret"},
```
HofmeisterAn, line 76: @galvesribeiro Is there a rule when a prefix should be used and when it should not? For example, some swarm types use a prefix while others do not. This situation is now similar to volume types. I have added the "Volume" prefix to these volume types to avoid clashes that would occur otherwise.

galvesribeiro: Yeah, never had a rule before. But in the Swarm case, would make sense that all of them had the Swarm prefix. You are good with the Volume prefix in this case, thanks.

PR #672, opened 2024-03-07
Migrate from Newtonsoft.Json to System.Text.Json

0xced:
Everything is now truly async, no more fake `Task.Factory.StartNew` async.


src/Docker.DotNet/Endpoints/StreamUtil.cs
```diff
@@ -29,16 +28,10 @@ internal static async Task MonitorStreamAsync(Task<Stream> streamTask, DockerCli
 
         internal static async Task MonitorStreamForMessagesAsync<T>(Task<Stream> streamTask, DockerClient client, CancellationToken cancellationToken, IProgress<T> progress)
         {
-            var tcs = new TaskCompletionSource<bool>();
-
             using (var stream = await streamTask)
-            using (var reader = new StreamReader(stream, new UTF8Encoding(false)))
-            using (var jsonReader = new JsonTextReader(reader) { SupportMultipleContent = true })
-            using (cancellationToken.Register(() => tcs.TrySetCanceled(cancellationToken)))
             {
```
dgvives, line 32: Verify this task can be cancelled by running the test `CreateContainerAsync_TimeoutExpires_Fails`.

src/Directory.Build.props
```diff
@@ -5,6 +5,7 @@
     <PropertyGroup>
         <IsPackable>true</IsPackable>
         <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
+        <LangVersion>latest</LangVersion>
```
paulomorgado, line 8: Isn't `latest` the default? What's the added benefit of this change?

paulomorgado:
> Also, in commit [6a87013](https://github.com/dotnet/Docker.DotNet/commit/6a87013e18ed0f02c9e296a53f1707e538e4780c) I improved all the HTTP requests to deserialize the JSON payload directly from the HTTP response stream instead of reading the response as string then deserializing to model objects. This is an improvement in both speed and memory allocations.

Why not just use [HttpClientJsonExtensions](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.json.httpclientjsonextensions) with [source generators](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation)?

I'm missing what the added value is for having generic methods with no constraints for making the HTTP requests.

src/Directory.Build.props
0xced: The default language version depends on the target framework, see the [C# language versioning](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-versioning#defaults) documentation.

For .NET Standard 2.0 the language version is C# 7.3 and for .NET Standard 2.1 the language version is C# 8.0.

I don't remember everything I changed but there's a high probability that I used some feature that requires C# greater than 7.3.

0xced:
> Why not just use [HttpClientJsonExtensions](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.json.httpclientjsonextensions) with [source generators](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation)?

I don't remember but it was probably because of the dozen of `MakeRequestAsync` variations that depend on `ApiResponseErrorHandlingDelegate` which whould have been a way too risky refactoring and also nearly impossible to review.

src/Directory.Build.props
paulomorgado: My mistake.

But it may be different for different SDK versions.

But, if you're aiming for a version higher than the supported for that target, you need to be very careful.

galvesribeiro:
It LGTM. For now, while we don't move toward the new code which drop the Go generator, this change to STJ will help others and yes, the extra allocations with strings will be dealt with in a separated PR. For now let's add it.

Also, the CI error is unrelated, so I'm good to push it as is. Will fix the CI separately. 

Thank you @0xced!

PR #683, opened 2024-08-19
Add copyUIDGID to ContainerPathStatParameters

macsux:
Allows extracting Tar balls with original owner. See spec [ContainerPathStatParameters](https://docs.docker.com/engine/api/v1.45/#tag/Container/operation/PutContainerArchive)


galvesribeiro (approved):
LGTM

Issue #663, opened 2023-12-11
MonitorEventsAsync Progress Action is one event behind


kosimas:
**Output of `dotnet --info`:**

```
.NET SDK:
 Version:           8.0.100
 Commit:            57efcf1350
 Workload version:  8.0.100-manifests.6a1e483a

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\8.0.100\

.NET workloads installed:
 Workload version: 8.0.100-manifests.6a1e483a
There are no installed workloads to display.

Host:
  Version:      8.0.0
  Architecture: x64
  Commit:       5535e31a71

.NET SDKs installed:
  6.0.320 [C:\Program Files\dotnet\sdk]
  6.0.417 [C:\Program Files\dotnet\sdk]
  7.0.404 [C:\Program Files\dotnet\sdk]
  8.0.100 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.25 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 7.0.14 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.0 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.NETCore.App 3.1.32 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 5.0.17 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 6.0.25 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 7.0.14 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.NETCore.App 8.0.0 [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
  Microsoft.WindowsDesktop.App 3.1.32 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
  Microsoft.WindowsDesktop.App 5.0.17 [C:\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App]
[...]

PR #692, opened 2025-02-09
Fixed CI build by skipping Bad Test

james-gould:
The `docker.exe` process is not always running, so this test fails due to luck/timing.

If `docker.exe` _wasn't_ running, the other tests would fail as the `TestFixture.DockerClient` would not be able to communicate with the local system's Docker instance.

### DIFF
--- a/src/Docker.DotNet/Models/RestartPolicyKind.cs
+++ b/src/Docker.DotNet/Models/RestartPolicyKind.cs
@@ -5,9 +5,6 @@ namespace Docker.DotNet.Models
 
     public enum RestartPolicyKind
     {
-        [EnumMember(Value = "")]
-        Undefined,
-
         [EnumMember(Value = "no")]
         No,
 
--- a/test/Docker.DotNet.Tests/Docker.DotNet.Tests.csproj
+++ b/test/Docker.DotNet.Tests/Docker.DotNet.Tests.csproj
@@ -1,11 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.12.0" />
     <PackageReference Include="xunit" Version="2.4.1" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.3"/>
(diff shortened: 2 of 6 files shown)


james-gould:
@dotnet-policy-service agree

james-gould:
The repo has had a lot of fixes and improvements since the last successful NuGet release on May 18th 2023. I ran into a bug raised as an issue in #663  and wanted to fix it, turns out the fix is already in place but the release hasn't gone out for a while.

src/Docker.DotNet/Models/RestartPolicyKind.cs
```diff
@@ -5,9 +5,6 @@ namespace Docker.DotNet.Models
 
     public enum RestartPolicyKind
     {
-        [EnumMember(Value = "")]
-        Undefined,
```
james-gould, line 9: Removed to mirror the updated type: https://github.com/moby/moby/blob/master/api/types/container/hostconfig.go#L283

Caused the updated System.Text.Json and System.Net.Http.Json serializers to fail due to empty strings not being valid.

All tests passing after this squash too.

test/Docker.DotNet.Tests/Docker.DotNet.Tests.csproj
```diff
@@ -1,11 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net8.0</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.2.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.12.0" />
```
james-gould, line 6: This resolved a critical vulnerability in the transitive Newtonsoft.Json dependency

james-gould:
Also concious of the version being behind, the current Docker API is v1.47 so the updated SEMVAR would be `3.147.0`. Not sure what the etiquette is for updating the version, if you'd like me to put that in the PR too let me know.

james-gould:
Not sure why my CI is failing on Github when locally the tests are working fine. Not familiar with Docker inside of a CI runner - any reason why this might be working locally but failing on Github? Even adding some trace logs reveals all looks fine - but it's fails.

https://github.com/james-gould/Docker.DotNet/actions/runs/13230419475

2025-02-09 23_06_29-Window

HofmeisterAn:
I would like to share that there is a [forked](https://github.com/testcontainers/Docker.DotNet/pull/8) version that resolves the issue without removing the enum value. According to the Docker Engine API documentation, `""` is a valid value.

james-gould:
Thanks Andre, didn't realise there was a maintained fork.  Will port my dep over and abandon this - clearly not my finest work!

PR #705, opened 2025-08-19
Fix leak with redirect

ybonnel:
When you call Inspect with a leading "/" in name (which is a value you have when getting name from a precedent inspect), docker engine first answer with a redirect.

Without the dispose, first request ends with a response not disposed.
On docker windows we ends with a leak of file handle on the named pipe.


HofmeisterAn:
FYI: This repo isn't very active anymore. We've [forked](https://github.com/testcontainers/Docker.DotNet) it and are adding important fixes and improvements there. We're also working on refactoring the API to be closer to how docker-java works. Your issue has already been [fixed](https://github.com/testcontainers/Docker.DotNet/blob/136a2870f16115d76f5ccebe9674ffe456db892e/src/Docker.DotNet/Microsoft.Net.Http.Client/ManagedHandler.cs#L75) in the forked version.

ybonnel:
@HofmeisterAn thx a lot for the information.

Will try to change my dep to your fork to see if I can migrate easilly.

PR #706, opened 2025-08-22
Support AOT

Egbert-Jan:


Egbert-Jan:
@dotnet-policy-service agree

HofmeisterAn:
Hi 👋, just a quick heads-up: This repo is pretty inactive. We've [forked](https://github.com/testcontainers/Docker.DotNet) it and are taking care of maintenance, bug fixes, and new features. If you'd like to make this PR compatible with our forked version, I'd be happy to review and merge it.

Here are a few things I noticed after a quick look:

- Doesn't it require project-level AOT configuration?
- Ideally, we should generate the `DockerDotnetJsonSerializerContext.Generated.cs` file using the Go implementation that already generates the model classes.
- Some things were removed, and I'm not sure why, but we can discuss that in a PR on the forked version.

Thanks for contributing, really appreciate the effort.

